PR #40, opened 2020-11-11
Add sync_complete() callback

### LINKED ISSUE #37: Add write() method to DiffSync class
### Environment
* DiffSync version:  1.0.0

Currently, to add bulk write operations to a DiffSync subclass object, one has to override the "sync_from" function on super(). This is in contrast to each individual DiffSyncModel's create(), update(), and delete() methods which are more idempotent in manner, and called by the "sync_from()" method on super.

### Proposed Functionality

It may be beneficial to be able to leave the downstream create(), update(), and delete() DiffSyncModel methods which are called by "sync_from" by default unimplemented and add a write() method to the DiffSync class. This would provide a framework for bulk write operations, without blowing away the logic implemented on "sync_from()"

### Use Case

class BackendYAML(DiffSync):
    def write(self, source):
        """Bulk write operation to dump data to disk from another backend
             Called automatically by super().sync_from()

        Args:
            source (DiffSync): DiffSync object from which data is being synchronized
        """
        # Validate whether or not any changes need to be made to the circuits/providers files
        self._write_providers_from(source)
        self._write_circuits_from(source)

glennmatthews:
Fixes #37, although I chose to name the callback method `sync_complete()` rather than `write()` as I think it's a bit clearer.

Note that this callback is not triggered if there are no actual diffs to synchronize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- #37 - add `sync_complete` callback, triggered on `sync_from` completion with changes.
+
 ## v1.0.0 - 2020-10-23
 
 Initial release
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 DiffSync is a utility library that can be used to compare and synchronize different datasets.
 
-For example, it can be used to compare a list of devices from 2 inventories system and, if required, synchronize them in either direction.
+For example, it can be used to compare a list of devices from 2 inventory systems and, if required, synchronize them in either direction.
 
 ```python
 A = DiffSyncSystemA()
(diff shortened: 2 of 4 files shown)


dgarros (approved):

PhillSimonds:
I think this looks good. Some food for thought...

I could see some confusion on knowing where to put logic for synchronizing in-memory objects vs where to put logic for synchronizing the backend. The same function (`create()`, `update()`, `delete()`) is used for both object synchronization and backend synchronization with one methodology, where if you're doing a bulk write to the backend, the logic is abstracted into it's own call (`sync_complete()`). I don't know that there's a better way to solve this, we probably just need to make sure it's explained well in the documentation.

glennmatthews:
Good points @PhillSimonds. Let me see what I can do as far as updates to the README.

PhillSimonds:
LGTM

dgarros (approved):

PR #42, opened 2020-11-18
Add "summary" method to Diff/DiffElement, fix an issue in DiffElement.dict()

### LINKED ISSUE #41: Diff should be able to provide a summary (# create/update/delete)
### Environment
* DiffSync version:  1.0.0

### Proposed Functionality

A Diff object should be able to provide a summary of its contents (i.e., the number of objects that would be created/updated/deleted if this diff were used for a synchronization between systems).

### Use Case

Logging, usability.

glennmatthews:
Fixes #41. An alternate approach to meeting this requirement that I considered but decided not to implement would be to have a method that could be called to generate logs for each object that a Diff/DiffElement *would* create/update/delete/leave unchanged, parallel to the logs generated during an actual sync by `DiffSync._sync_from_diff_element()`. **I'm open to the argument that it would be worthwhile to implement that option as well - feedback would be appreciated!**

In the process of refactoring some of the unit tests to reduce repetition, I encountered an issue that the `DiffElement.dict()` method return value didn't distinguish between "source and target both exist but have no relevant attributes" and "source and/or target does not exist at all" - both would be the same dict `{"+": {}, "-": {}}`. I've therefore altered the behavior of this method (and the associated `get_attrs_diffs()` method) so that a completely absent source/dest results in the "+" or "-" key being omitted altogether, making the distinction more clear.

### DIFF
--- a/diffsync/diff.py
+++ b/diffsync/diff.py
@@ -101,6 +101,7 @@ class Diff:
             "create": 0,
             "update": 0,
             "delete": 0,
+            "no-change": 0,
         }
         for child in self.get_children():
             child_summary = child.summary()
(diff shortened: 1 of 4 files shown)


PhillSimonds (approved):
LGTM

dgarros (approved):
Looks great

PR #43, opened 2020-11-18
get_by_uids() should raise ObjectNotFound when appropriate

glennmatthews:
Instead of silently skipping over unmatched uids, `get_by_uids()` now raises an `ObjectNotFound` exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


dgarros (dismissed):
I think that make sense
Should we raise the same exception in `get()` to be consistent ?

glennmatthews:
Could do. At least there we're currently returning `None` rather than silently ignoring the failure, but I think you're right that raising an exception by default would be safer.

dgarros (approved):
LGTM, thanks for adding get as well

PR #45, opened 2020-12-05
Update project to support Python 3.6 & 3.9 and update travis to use stages

dgarros:
This PR update pyproject.toml and the dependencies to support Python 3.6 and python 3.9.
No changes in the code were required.

Also in the process, I did some cleanup
- Updated some files to align with the latest dev-standard
- Removed unused deps like request_mocks and ipython
 - Refactor the travis file with stages, leveraging the native python version in travis.


glennmatthews:
Looks like something isn't quite right with the pyproject.toml changes versus the poetry.lock changes:

```
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.
```

tasks.py
```diff
@@ -15,263 +15,269 @@
...
-        python_ver (str): Will use the Python version docker image to build from
-    """
-    # pty is set to true to properly run the docker commands due to the invocation process of docker
-    # https://docs.pyinvoke.org/en/latest/api/runners.html - Search for pty for more information
-    docker = f"docker run -it -v {PWD}:/local {name}-{python_ver}:latest"
-    context.run(f"{docker} flake8 .", pty=True)
-
-
-@task
-def mypy(context, name=NAME, python_ver=PYTHON_VER):
```
glennmatthews, line 163: Looks like the `mypy` task got removed - please add it back.

tasks.py
```diff
@@ -15,263 +15,269 @@
...
     # pty is set to true to properly run the docker commands due to the invocation process of docker
     # https://docs.pyinvoke.org/en/latest/api/runners.html - Search for pty for more information
     # Install python module
-    docker = f"docker run -it -v {PWD}:/local {name}-{python_ver}:latest"
-    context.run(
-        f"{docker} /bin/bash -c 'poetry install && pytest "
-        "--cov=diffsync --cov-config pyproject.toml --cov-report html --cov-report term -vv'",
-        pty=True,
-    )
+    exec_cmd = "pytest -vv"
```
glennmatthews, line 152: Looks like you dropped the use of `coverage` here - intentional?

tasks.py
dgarros: good catch, I added it back

tasks.py
dgarros: good catch, I added it back, sorry about that

.travis.yml
```diff
@@ -1,28 +1,67 @@
...
+      before_script:
+        - "pip install invoke poetry toml"
+        - "poetry install --no-interaction --no-ansi --no-root"
+      script:
+        - "invoke black"
+        - "invoke bandit"  # Bandit fails to function on > Py3.8 https://github.com/PyCQA/bandit/issues/639
+        - "invoke pydocstyle"
+        - "invoke flake8"
+        - "invoke yamllint"
+        - "invoke pylint"
```
glennmatthews, line 47: Need `invoke mypy` here as well.

tasks.py
```diff
@@ -15,263 +15,284 @@
 limitations under the License.
 """
 import os
-from invoke import task  # type: ignore
```
glennmatthews, line 18: Does mypy no longer require the `type: ignore` comment to avoid complaining about invoke?

glennmatthews (approved):
One more mypy fix needed, otherwise looks good to me. Thanks!

glennmatthews (approved):

PR #48, opened 2021-04-05
Add progress callback and example

### LINKED ISSUE #47: Add option for progress-bar type reporting
### Environment
* DiffSync version: 1.2.0

### Proposed Functionality

DiffSync to provide hooks for status reporting, for example:
- number of records processed so far in diff calculation
- number of records remaining for diff calculation
- number of diff elements processed so far in synchronization
- number of diff elements remaining to process for synchronization

Example API to consider is that of urllib's `reporthook` (https://docs.python.org/3/library/urllib.request.html#urllib.request.URLopener.retrieve):

> If reporthook is given, it must be a function accepting three numeric parameters: A chunk number, the maximum size chunks are read in and the total size of the download (-1 if unknown). It will be called once at the start and after each chunk of data is read from the network.

### Use Case

When using DiffSync for a large set of records, both diffing and syncing may take some time to complete. Although logging can be enabled to get highly detailed information about DiffSync's progress, it would be useful to have a less-detailed status/progress information API as well, which could be used to (for example) update a progress bar.

glennmatthews:
Fixes #47.

Added optional `callback` argument to `diff_from/diff_to/sync_from/sync_to` APIs. This allows the caller to provide a callback function which will be invoked perioidically as `callback(stage, current, total)` during the diff and sync processes, which can be used to print status to the console or to update a progress bar - useful especially during long-running diffs and syncs of large data sets such as an [import from NetBox to Nautobot](https://github.com/nautobot/nautobot-plugin-netbox-importer/pull/23).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,11 @@ A.sync_from(B)
 A.sync_to(B)
 ```
 
+You may wish to peruse the [`diffsync` GitHub topic](https://github.com/topics/diffsync) for examples of projects using this library.
+
 # Getting started
 
-To be able to properly compare different datasets, DiffSync relies on a shared datamodel that both systems must use.
+To be able to properly compare different datasets, DiffSync relies on a shared data model that both systems must use.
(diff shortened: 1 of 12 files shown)


dgarros (approved):
LGTM

Issue #52, opened 2021-06-03
Docs Update: Clarify sublassing DiffSyncModels for different backends


tyler-8:
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
The README makes mention of extending a "base" DiffSyncModel for handling CRUD actions in a backend, but doesn't do a great job of visualizing this concept.

> you need to extend your DiffSyncModel class(es) to define your own create, update and/or delete methods for each model.

I think extending the example out a little more would go a long way to showing how to build your models and adapters.

```python
class Device(DiffSyncModel):
    """Example model of a network Device."""
​
    _modelname = "device"
    _identifiers = ("name",)
    _attributes = ()
    _children = {"interface": "interfaces"}
​
    name: str
    site_name: Optional[str]  # note that this attribute is NOT included in _attributes
    role: Optional[str]  # note that this attribute is NOT included in _attributes
    interfaces: List = list()
​
​
class SystemADevice(Device):

    system_A_unique_field: Optional[str] = None
    
    @classmethod
    def create(cls, diffsync, ids, attrs):
        """Talk to SystemA to create device"""
        pass
​
class SystemBDevice(Device):
    
    system_B_unique_field: Optional[str] = None

    @classmethod
    def create(cls, diffsync, ids, attrs):
        """Talk to SystemB to create device"""
        pass
```

### Use Case

This should help newbies (like myself) to get a better idea of how to architect a diffsync-based integration.

dgarros:
I've created a new example that partially address this issue #53 but nonetheless we should plan to have a dedicated section of the doc on this

PR #53, opened 2021-06-18
Add new example with pynautobot

dgarros:
This PR adds a new example to show how to use DiffSync to interact with a remote system via its API, like Nautobot.

### DIFF
--- a/examples/example3/README.md
+++ b/examples/example3/README.md
@@ -1,0 +1,32 @@
+
+# Example 3
+
+This is a simple example to show how DiffSync can be used to compare and synchronize data with a remote system like via a REST API like Nautobot.
+
+For this example, we have a shared model for Region and Country defined in `models.py`.
+A Country must be associated with a Region and can be part of a Subregion too.
+
+The comparison and synchronization of dataset is done between a local JSON file and the [public instance of Nautobot](https://demo.nautobot.com).
+
+
+## Install the requirements
+
+to use this example you must have some dependencies installed, please make sure to run 
+```
+pip install -r requirements.txt
+```
+
+## Try the example
+
+The first time a lot of changes should be reported between Nautobot and the local data because by default the demo instance doesn't have the subregion define.
+After the first sync, the diff should show no difference. 
+At this point, Diffsync will be able to identify and fix all changes in Nautobot. You can try to add/update or delete any country in Nautobot and DiffSync will automatically catch it and it will fix it with running in sync mode.
+
+```
+### DIFF Compare the data between Nautobot and the local JSON file.
+main.py --diff
+
+### SYNC Update the list of country in Nautobot.
+main.py --sync
+```
+
--- a/examples/example3/local_adapter.py
+++ b/examples/example3/local_adapter.py
@@ -1,0 +1,55 @@
+import json
+
+from slugify import slugify
+
+from diffsync import DiffSync
+from models import Region, Country
+
+COUNTRIES_FILE = "countries.json"
+
+
+class LocalAdapter(DiffSync):
+    """DiffSync Adapter to Load the list of regions and countries from a local JSON file."""
+
+    region = Region
+    country = Country
+
+    # Since all countries are associated with a region, we don't need to list country here
+    # When doing a diff or a sync between 2 adapters,
+    #  diffsync will recursively check all models defined at the top level and their children.
+    top_level = ["region"]
+
+    # Human readable name of the Adapter,
+    # mainly used when doing a diff to indicate where each data is coming from
+    type = "Local"
+
+    def load(self, filename=COUNTRIES_FILE):
+        """Load all regions and countries from a local JSON file."""
+
+        data_file = open(filename, "r")
+        countries = json.loads(data_file.read())
+
+        # Load all regions first
+        # A Region object will be create for each region and it will be store inside the object with self.add
+        # To create a Region we are using "self.region" instead of "Region" directly to allow someone to extend this adapter without redefining everything.
+        region_names = set([country.get("region") for country in countries])
+        for region in region_names:
+            self.add(self.region(slug=slugify(region), name=region))
+
+        # Load all countries
+        # A Country object will be create for each country, it will be store inside the object with self.add
+        # and it will be linked to its parent with parent.add_child(item)
+        for country in countries:
+
+            # Retrive the parent region object from the internal cache.
+            region = self.get(obj=self.region, identifier=slugify(country.get("region")))
+
+            name = country.get("country")
+            item = self.country(
+                slug=slugify(name), name=name, subregion=country.get("subregion", None), region=region.slug
+            )
+            self.add(item)
+
+            region.add_child(item)
+
+        data_file.close()
--- a/examples/example3/main.py
+++ b/examples/example3/main.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+"""Main executable for DiffSync "example2"."""
+import sys
+import argparse
+import pprint
+
+from diffsync import Diff
+from diffsync.logging import enable_console_logging
+
+from local_adapter import LocalAdapter
+from nautobot_adapter import NautobotAdapter
+
+
+def main():
+    """Demonstrate DiffSync behavior using the example backends provided."""
+    parser = argparse.ArgumentParser("example1")
+    parser.add_argument("--verbosity", "-v", default=0, action="count")
+    parser.add_argument("--diff", action="store_true")
+    parser.add_argument("--sync", action="store_true")
+    args = parser.parse_args()
+    enable_console_logging(verbosity=args.verbosity)
+
+    if not args.sync and not args.diff:
+        sys.exit("please select --diff or --sync")
+
+    print("Initializing and loading Local Data ...")
+    local = LocalAdapter()
+    local.load()
+    # print(local.str())
+
+    print("Initializing and loading Nautobot Data ...")
+    nautobot = NautobotAdapter()
+    nautobot.load()
+    # print(nautobot.str())
+
+    if args.diff:
+        print("Calculating the Diff between the local adapter and Nautobot ...")
+        diff = nautobot.diff_from(local)
+        print(diff.str())
+
+    elif args.sync:
+        print("Updating the list of countries in Nautobot ...")
+        nautobot.sync_from(local)
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/example3/models.py
+++ b/examples/example3/models.py
@@ -1,0 +1,34 @@
+from typing import List, Optional
+from diffsync import DiffSyncModel
+
+
+class Region(DiffSyncModel):
+    """Example model of a geographic region."""
+
+    _modelname = "region"
+    _identifiers = ("slug",)
+    _attributes = ("name",)
+
+    # By listing country as a child to Region
+    # DiffSync will be able to recursively compare all regions including all their children
+    _children = {"country": "countries"}
+
+    slug: str
+    name: str
+    countries: List[str] = list()
+
+
+class Country(DiffSyncModel):
+    """Example model of a Country.
+
+    A must be part of a region and can be also associated with a subregion.
+    """
+
+    _modelname = "country"
+    _identifiers = ("slug",)
+    _attributes = ("name", "region", "subregion")
+
+    slug: str
+    name: str
+    region: str
+    subregion: Optional[str]
--- a/examples/example3/nautobot_adapter.py
+++ b/examples/example3/nautobot_adapter.py
@@ -1,0 +1,167 @@
+import os
+import pynautobot
+
+from diffsync import DiffSync
+from models import Region, Country
+
+NAUTOBOT_URL = os.getenv("NAUTOBOT_URL", "https://demo.nautobot.com")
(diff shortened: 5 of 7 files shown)


examples/example3/README.md
```diff
@@ -0,0 +1,32 @@
+
+# Example 3
+
+This is a simple example to show how DiffSync can be used to compare and synchronize data with a remote system like via a REST API like Nautobot.
```
glennmatthews, line 4:
```suggestion
This is a simple example to show how DiffSync can be used to compare and synchronize data with a remote system like [Nautobot](https://nautobot.readthedocs.io) via a REST API.
```

examples/example3/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+For this example, we have a shared model for Region and Country defined in `models.py`.
+A Country must be associated with a Region and can be part of a Subregion too.
```
glennmatthews, line 7: I'm not sure the Subregion adds much to this example, and I find it a bit confusing - it looks like a "Subregion" is just used as the `description` of the constructed Country `region`? But Nautobot supports nesting of `region`s, so why not assign the Country directly to the most relevant "sub-"`region` instead of the more general base `Region`?

examples/example3/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Install the requirements
+
+to use this example you must have some dependencies installed, please make sure to run 
+```
+pip install -r requirements.txt
+```
+
+## Try the example
+
+The first time a lot of changes should be reported between Nautobot and the local data because by default the demo instance doesn't have the subregion define.
```
glennmatthews, line 21:
```suggestion
The first time you run this example, a lot of changes should be reported between Nautobot and the local data because by default the demo instance doesn't have the subregion defined.
```

Would be good to include a snippet of sample output here as well.

examples/example3/README.md
```diff
@@ -0,0 +1,32 @@
...
+After the first sync, the diff should show no difference.
```
glennmatthews, line 22:
```suggestion
After the first sync, on subsequent runs, the diff should show no changes. 
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+
+class LocalAdapter(DiffSync):
+    """DiffSync Adapter to Load the list of regions and countries from a local JSON file."""
+
+    region = Region
+    country = Country
+
+    # Since all countries are associated with a region, we don't need to list country here
+    # When doing a diff or a sync between 2 adapters,
+    #  diffsync will recursively check all models defined at the top level and their children.
```
glennmatthews, line 19:
```suggestion
    # When doing a diff or a sync between 2 adapters,
    #  DiffSync will recursively check all models defined at the top level and their children.
    # Since our data model defines countries as children of a region, we don't need to list country here
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+from diffsync import DiffSync
+from models import Region, Country
+
+COUNTRIES_FILE = "countries.json"
+
+
+class LocalAdapter(DiffSync):
+    """DiffSync Adapter to Load the list of regions and countries from a local JSON file."""
+
+    region = Region
```
glennmatthews, line 14:
```suggestion
    # Define all data models that this adapter makes use of.
    # Note that the variable names ("region", "country") need to match between DiffSync Adapter classes.
    region = Region
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # Human readable name of the Adapter,
+    # mainly used when doing a diff to indicate where each data is coming from
+    type = "Local"
+
+    def load(self, filename=COUNTRIES_FILE):
+        """Load all regions and countries from a local JSON file."""
+
+        data_file = open(filename, "r")
+        countries = json.loads(data_file.read())
```
glennmatthews, line 30:
```suggestion
        with open(filename, "r") as data_file:
            countries = json.load(data_file)
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # Load all regions first
+        # A Region object will be create for each region and it will be store inside the object with self.add
```
glennmatthews, line 33:
```suggestion
        # A Region object will be created for each region and it will be stored inside the adapter with self.add()
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+        # To create a Region we are using "self.region" instead of "Region" directly to allow someone to extend this adapter without redefining everything.
+        region_names = set([country.get("region") for country in countries])
+        for region in region_names:
+            self.add(self.region(slug=slugify(region), name=region))
+
+        # Load all countries
+        # A Country object will be create for each country, it will be store inside the object with self.add
```
glennmatthews, line 40:
```suggestion
        # A Country object will be created for each country, it will be stored inside the adapter with self.add(),
```

examples/example3/local_adapter.py
```diff
@@ -0,0 +1,55 @@
...
+
+            name = country.get("country")
+            item = self.country(
+                slug=slugify(name), name=name, subregion=country.get("subregion", None), region=region.slug
+            )
+            self.add(item)
+
+            region.add_child(item)
+
+        data_file.close()
```
glennmatthews, line 55: Can be removed if you're using the "with open()..." context manager as suggested above.

```suggestion
```

examples/example3/main.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+"""Main executable for DiffSync "example2"."""
```
glennmatthews, line 2:
```suggestion
"""Main executable for DiffSync "example3"."""
```

examples/example3/main.py
```diff
@@ -0,0 +1,47 @@
...
+from diffsync import Diff
+from diffsync.logging import enable_console_logging
+
+from local_adapter import LocalAdapter
+from nautobot_adapter import NautobotAdapter
+
+
+def main():
+    """Demonstrate DiffSync behavior using the example backends provided."""
+    parser = argparse.ArgumentParser("example1")
```
glennmatthews, line 16:
```suggestion
    parser = argparse.ArgumentParser("example3")
```

examples/example3/main.py
```diff
@@ -0,0 +1,47 @@
...
+    nautobot = NautobotAdapter()
+    nautobot.load()
+    # print(nautobot.str())
+
+    if args.diff:
+        print("Calculating the Diff between the local adapter and Nautobot ...")
+        diff = nautobot.diff_from(local)
+        print(diff.str())
+
+    elif args.sync:
```
glennmatthews, line 41: It doesn't look like `--diff` and `--sync` are mutually exclusive arguments

```suggestion
    if args.sync:
```

examples/example3/main.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+"""Main executable for DiffSync "example2"."""
+import sys
+import argparse
+import pprint
```
glennmatthews, line 5: Appears unused

```suggestion
```

examples/example3/models.py
```diff
@@ -0,0 +1,34 @@
+from typing import List, Optional
```
glennmatthews, line 1: Consider renaming this file to `local_models.py` and extracting the Nautobot-specific models from `nautobot_adapter.py` into a `nautobot_models.py` file? I think this would be a good step in the direction of addressing #52.

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
+import os
+import pynautobot
+
+from diffsync import DiffSync
+from models import Region, Country
+
+NAUTOBOT_URL = os.getenv("NAUTOBOT_URL", "https://demo.nautobot.com")
+NAUTOBOT_TOKEN = os.getenv("NAUTOBOT_TOKEN", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
```
glennmatthews, line 8: Might be a good idea to add a note in the README about these two environment variables and how you can set them if you want to point to a local instance of Nautobot instead of the public demo instance?

examples/example3/main.py
```diff
@@ -0,0 +1,47 @@
...
+    # print(nautobot.str())
+
+    if args.diff:
+        print("Calculating the Diff between the local adapter and Nautobot ...")
+        diff = nautobot.diff_from(local)
+        print(diff.str())
+
+    elif args.sync:
+        print("Updating the list of countries in Nautobot ...")
+        nautobot.sync_from(local)
```
glennmatthews, line 43: Possibly include (at least as an option) the `IGNORE_UNMATCHED_DST` flag so that Regions in Nautobot that don't exist in the local file (such as `networktocode`, but possibly including others as well) do not get deleted?

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+
+class NautobotRegion(Region):
+    """Extend the Region object to store Nautobot specific information.
+
+    Region are represented in Nautobot as a dcim.region object without parent.
+    """
+
+    remote_id: str
+    """Store the nautobot uuid in the object to allow update and delete of existing object."""
+
```
glennmatthews, line 19: This reminds me that we need a pattern for models that do not support CRUD operations - right now it's more than a bit confusing as DiffSync will log messages *saying* that it's performing create/update/delete, it's just actually a no-op that isn't actually changing any underlying system. 

Maybe for now create stub CRUD methods on this model that just log a message saying "Nautobot is not being updated"?

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    
+    Country are represented in Nautobot as a dcim.region object as well but a country must have a parent.
+    Subregion information will be store in the description of the object in Nautobot
+    """
+
+    remote_id: str
+    """Store the nautobot uuid in the object to allow update and delete of existing object."""
+
+    @classmethod
+    def create(cls, diffsync: "DiffSync", ids: dict, attrs: dict):
```
glennmatthews, line 32:
```suggestion
    def create(cls, diffsync: DiffSync, ids: dict, attrs: dict):
```

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+        Args:
+            diffsync: The master data store for other DiffSyncModel instances that we might need to reference
+            ids: Dictionary of unique-identifiers needed to create the new object
+            attrs: Dictionary of additional attributes to set on the new object
+
+        Returns:
+            NautobotCountry: DiffSync object newly created
+        """
+
+        # Retrieve the parent region in internal cache to access its UUID
```
glennmatthews, line 44:
```suggestion
        # Retrieve the parent region from the adapter's internal cache to access its UUID
```

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+        # Retrieve the pynautobot object and delete the object in Nautobot
+        remote = self.diffsync.nautobot.dcim.regions.get(self.remote_id)
+        remote.delete()
+
+        super().delete()
+        return self
+
+
+class NautobotAdapter(DiffSync):
+    """Example of a DiffSync adapter implementation."""
```
glennmatthews, line 112:
```suggestion
    """Example of a DiffSync adapter implementation using pynautobot to communicate with a remote Nautobot system."""
```

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # We are using NautobotCountry and NautobotRegion instead of Region and Country
+    # because we are using these classes to manage the logic to integrate with Nautobot
+    # NautobotRegion is just a small extension to store the UUID and do not support any CRUD operation toward Nautobot
```
glennmatthews, line 116:
```suggestion
    # NautobotRegion is just a small extension to store the UUID and does not support any CRUD operation toward Nautobot
```

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    # NautobotCountry support the creation, update or deletion of a country in Nautobot
```
glennmatthews, line 117:
```suggestion
    # NautobotCountry supports the creation, update or deletion of a country in Nautobot
```

examples/example3/nautobot_adapter.py
```diff
@@ -0,0 +1,167 @@
...
+    region = NautobotRegion
+    country = NautobotCountry
+
+    # Since all countries are associated with a region, we don't need to list country here
+    # When doing a diff or a sync between 2 adapters,
+    #  diffsync will recursively check all models defined at the top level and their children.
```
glennmatthews, line 123:
```suggestion
    # When doing a diff or a sync between 2 adapters,
    #  diffsync will recursively check all models defined at the top level and their children.
    # Since countries are defined as children of a region, we don't need to list country here
```

examples/example3/README.md
dgarros: Somewhat agreed, I was just looking for a field to update that is not part of the identifier ...  
I'll update the example to create a custom field under dcim.site to store the population per country

glennmatthews: Sounds good, I think that'll be clearer. Thanks!

Issue #57, opened 2021-06-20
Support multiple types of datastore 


dgarros:
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
Currently DiffSync Adapters are always leveraging an internal in-memory datastore that is storing the entire dataset. 
It would be great to support different types of datastore, like Redis in addition to the in-memory datastore.
As an option it would be useful to deactivate the internal datastore as well or provide a solution to pull the data directly from  the remote system.

### Use Case
When we are dealing with a large dataset, the volume of data stored in-memory can become very large and can present some challenges. And external datastore like Redis would reduce the volume of data stored in memory.
In some cases, DiffSync is running very closely to an existing database and duplicating the data in memory is redundant and inefficient.

chadell:
First abstraction and Redis support done in #106

PR #58, opened 2021-06-20
Prototype/example of an adapter that doesn't leverage the internal datastore

dgarros:
This PR is a prototype to explore if it's possible to build an adapter that is not leveraging the internal datastore (related to #57)

For this example I modified example3 (still in review #53) to bypass the internal datastore by reimplementing the 5 main methods on the adapter that are interacting with the datastore (`get` / `get_all` / `get_by_uids` / `add` /  `remove`).
I moved the logic of `get` / `get_all` and `get_by_uids` to each model class instead. 

it's working but it's significantly slower, especially when doing a sync with a very small diff, I need to investigate further why but it's a good start.
One thing I noticed is that DiffSync is currently expecting `get_all` and `get_by_uids` to return a list where it would be useful to accept of generator instead to improve the processing and the memory consumption. 

I'd love to get some feedback on this experiment and how we could package that into DiffSync if we want to move forward.

### DIFF
--- a/examples/example3/README.md
+++ b/examples/example3/README.md
@@ -1,0 +1,32 @@
+
+# Example 3
+
+This is a simple example to show how DiffSync can be used to compare and synchronize data with a remote system like via a REST API like Nautobot.
+
+For this example, we have a shared model for Region and Country defined in `models.py`.
+A Country must be associated with a Region and can be part of a Subregion too.
+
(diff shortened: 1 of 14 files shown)


glennmatthews (commented):
Interesting design! Definitely worth exploring further. What issues did you run into when trying to return a generator rather than a list from `get_all` and `get_by_uids`? Have you done any sort of memory profiling to see what sort of difference is observed?

chadell:
superseded by #106

PR #70, opened 2021-10-11
Add add_or_update to DiffSync class.

### LINKED ISSUE #50: Get or create function
### Environment
* DiffSync version:  1.1.0

### Ability to either get or create an object like the Django ORM

Function that takes the identifiers of an object and either gets or create an object.
Current work around is wrap an object get in an exception of `ObjectNotFound`.

### Use Case

Currently using `try` and `except` to catch if an object already exists. 

``` python
            try:
                self.add(vrf)
            except ObjectAlreadyExists:
                pass
```

Expected use
``` python
self.get_or_create(vrf)
self.update_or_create(vrf)
```

FragmentedPacket:
Fixes #50 

- Adds `add_or_update` to DiffSync class that requires a `DiffSyncModel` to be passed in and will attempt to add or update an existing object
- Convert one of the examples over to use the new method
- Add test to validate proper functionality


diffsync/__init__.py
```diff
@@ -692,6 +696,31 @@ def remove(self, obj: DiffSyncModel, remove_children: bool = False):
                         # Since this is "cleanup" code, log an error and continue, instead of letting the exception raise
                         self._log.error(f"Unable to remove child {child_id} of {modelname} {uid} - not found!")
 
+    def get_or_create(
```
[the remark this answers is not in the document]
FragmentedPacket: Correct. As we discussed, will change to `get_or_instantiate`.

diffsync/__init__.py
```diff
@@ -653,8 +653,12 @@ def add(self, obj: DiffSyncModel):
         modelname = obj.get_type()
         uid = obj.get_unique_id()
 
-        if uid in self._data[modelname]:
-            raise ObjectAlreadyExists(f"Object {uid} already present")
+        existing_obj = self._data[modelname].get(uid)
+        if existing_obj:
+            if existing_obj != obj:
+                raise ObjectAlreadyExists(f"Object {uid} already present")
```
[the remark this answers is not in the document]
FragmentedPacket: This will be in the next commit. I'm just passing in the object. Let me know if that isn't quite what you're looking for.

examples/01-multiple-data-sources/backend_b.py
```diff
@@ -49,15 +49,16 @@ class BackendB(DiffSync):
...
+                attrs = {"role": device_data["role"], "site_name": site_name}
+                device, _ = self.get_or_create(self.device, identifiers, attrs)
                 site.add_child(device)
 
                 for intf_name, desc in device_data["interfaces"].items():
-                    intf = self.interface(name=intf_name, device_name=device_name, description=desc)
-                    self.add(intf)
+                    identifiers = {"name": intf_name, "device_name": device_name}
+                    attrs = {"description": desc}
+                    intf, _ = self.get_or_create(self.interface, identifiers, attrs)
```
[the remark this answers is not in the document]
FragmentedPacket: I added an example to my latest commit. Now that I can see the example, would it be best to change these to actually add the objects as well rather than having the backend have to check after each call to these new methods?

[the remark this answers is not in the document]
FragmentedPacket: Spoke with Damien and he agreed we should add the new models to the DiffSync adapter. I just pushed up some updates and renaming per Damien (had the old names of the functions)

diffsync/__init__.py
```diff
@@ -648,13 +648,17 @@ def add(self, obj: DiffSyncModel):
             obj (DiffSyncModel): Object to store
 
         Raises:
-            ObjectAlreadyExists: if an object with the same uid is already present
+            ObjectAlreadyExists: if an object with the same uid and different attributes is already present.
```
glennmatthews, line 651:
```suggestion
            ObjectAlreadyExists: if a different object with the same uid is already present.
```

diffsync/__init__.py
```diff
@@ -692,6 +696,66 @@ def remove(self, obj: DiffSyncModel, remove_children: bool = False):
...
+            self.add(obj)
+            created = True
+
+        # Update existing obj with attrs
+        if attrs:
+            existing_attrs = obj.get_attrs()
+            mutual_attrs = set(existing_attrs).intersection(attrs)
+            for attr in mutual_attrs:
+                if existing_attrs[attr] != attrs[attr]:
+                    setattr(obj, attr, attrs[attr])
```
glennmatthews, line 755: Is the idea here to ignore any `attrs` that aren't in the object model's base attributes? That seems odd to me. I would have expected simply:

```suggestion
        for attr, value in attrs.items():
            if getattr(obj, attr) != value:
                setattr(obj, attr, value)
```

examples/04-get-update-instantiate/backends.py
```diff
@@ -0,0 +1,135 @@
...
+    def load(self):
+        """Initialize the BackendA Object by loading some site, device and interfaces from DATA."""
+        for device_data in BACKEND_DATA_A:
+            device, instantiated = self.get_or_instantiate(
+                self.device, {"name": device_data["name"]}, {"role": device_data["role"]}
+            )
+
+            site, instantiated = self.get_or_instantiate(self.site, {"name": device_data["site"]})
+            if instantiated:
+                device.add_child(site)
```
glennmatthews, line 98: It's a bit odd to me conceptually to have `site` as a child of `device` (I would have expected the opposite) but maybe it's fine for the sake of example?

tests/unit/test_diffsync.py
```diff
@@ -81,11 +81,126 @@ def test_diffsync_get_by_uids_with_no_data(generic_diffsync):
...
+    person = PersonA(name="Mikhail Yohman")
+
+    modelname = person.get_type()
+    uid = person.get_unique_id()
+
+    # First attempt at adding object
+    generic_diffsync.add(person)
+    assert modelname in generic_diffsync._data  # pylint: disable=protected-access
+    assert uid in generic_diffsync._data[modelname]  # pylint: disable=protected-access
+    assert person == generic_diffsync._data[modelname][uid]  # pylint: disable=protected-access
```
glennmatthews, line 94:
```suggestion
    assert person is generic_diffsync._data[modelname][uid]  # pylint: disable=protected-access
```

tests/unit/test_diffsync.py
```diff
@@ -81,11 +81,126 @@ def test_diffsync_get_by_uids_with_no_data(generic_diffsync):
...
+
+    # Attempt to add again and make sure it doesn't raise an exception
+    generic_diffsync.add(person)
+    assert person == generic_diffsync._data[modelname][uid]  # pylint: disable=protected-access
+    assert person == generic_diffsync.get(PersonA, "Mikhail Yohman")
```
glennmatthews, line 99:
```suggestion
    assert person is generic_diffsync._data[modelname][uid]  # pylint: disable=protected-access
    assert person is generic_diffsync.get(PersonA, "Mikhail Yohman")
```

tests/unit/test_diffsync.py
```diff
@@ -81,11 +81,126 @@ def test_diffsync_get_by_uids_with_no_data(generic_diffsync):
...
+    with pytest.raises(ObjectAlreadyExists) as error:
+        generic_diffsync.add(new_intf)
+    error_model = error.value.args[1]
+    assert isinstance(error_model, DiffSyncModel)
+    assert new_intf is error_model
+
+
+def test_diffsync_get_or_instantiate_create_non_existent_object(generic_diffsync):
+    intf_identifiers = {"device_name": "device1", "name": "eth1"}
+    intf = Interface(**intf_identifiers)
```
glennmatthews, line 117: Doesn't seem like `intf` is really needed for this test?

tests/unit/test_diffsync.py
```diff
@@ -81,11 +81,126 @@ def test_diffsync_get_by_uids_with_no_data(generic_diffsync):
...
+def test_diffsync_get_or_instantiate_retrieve_existing_object_wo_attrs(generic_diffsync):
+    intf_identifiers = {"device_name": "device1", "name": "eth1"}
+    intf = Interface(**intf_identifiers)
+    generic_diffsync.add(intf)
+
+    obj, created = generic_diffsync.get_or_instantiate(Interface, intf_identifiers)
+    assert obj is intf
+    assert not created
+    assert obj.interface_type == "ethernet"
+    assert obj.description is None
```
glennmatthews, line 159: Would it make sense to also have a test case where you have an existing object, then call `get_or_instantiate` with matching `ids` but with `attrs` that differ from the existing object, and assert that the retrieved object preserves the original object's `attrs`?

tests/unit/test_diffsync.py
```diff
@@ -81,11 +81,126 @@ def test_diffsync_get_by_uids_with_no_data(generic_diffsync):
...
+def test_diffsync_update_or_instantiate_retrieve_existing_object(generic_diffsync):
+    intf_identifiers = {"device_name": "device1", "name": "eth1"}
+    intf = Interface(**intf_identifiers)
+    generic_diffsync.add(intf)
+
+    obj, created = generic_diffsync.update_or_instantiate(Interface, intf_identifiers)
+    assert obj is intf
+    assert not created
+    assert obj.interface_type == "ethernet"
+    assert obj.description is None
```
glennmatthews, line 171: This looks like an exact duplicate of the previous test case to me - intentional?

diffsync/__init__.py
```diff
@@ -329,7 +329,7 @@ def add_child(self, child: "DiffSyncModel"):
         attr_name = self._children[child_type]
         childs = getattr(self, attr_name)
         if child.get_unique_id() in childs:
-            raise ObjectAlreadyExists(f"Already storing a {child_type} with unique_id {child.get_unique_id()}")
+            raise ObjectAlreadyExists(f"Already storing a {child_type} with unique_id {child.get_unique_id()}", child)
```
glennmatthews, line 332: It'd be nice to change the `__init__` method of `ObjectAlreadyExists` to make the "existing object" arg mandatory and cache it in a more useful location than `self.value.args[1]`, something like:

```python
class ObjectAlreadyExists(ObjectStoreException):
    """Exception raised when trying to store a DiffSyncModel or DiffElement that is already being stored."""

    def __init__(self, message, existing_object, *args, **kwargs):
        self.existing_object = existing_object
        super().__init__(message, existing_object, *args, **kwargs)
```

Then your tests can check for `error.existing_object` instead of `error.value.args[1]`.

examples/04-get-update-instantiate/backends.py
FragmentedPacket: Yeah it was more or less just for the sake of the example.

diffsync/__init__.py
FragmentedPacket: Do we want to consume the the `AttributeError` if the `obj` doesn't have the necessary attr and log it or raise a new custom error? Or should we default within the `getattr` to `None` if the attr doesn't exist?

examples/04-get-update-instantiate/backends.py
```diff
@@ -0,0 +1,135 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+# pylint: disable=wrong-import-order
+from diffsync import DiffSync
+from models import Site, Device, Interface  # pylint: disable=no-name-in-module
```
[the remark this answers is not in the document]
FragmentedPacket: Yes. Sorry I took these from existing examples. Changing the order fixes the `wrong-import-order`, and I removed the `no-name-in-module` without any other changes and pylint passes so not sure why they were there in the first place.

diffsync/__init__.py
FragmentedPacket: Ok I updated it to do this, but we do have to access it via `error.value.existing_object`.

diffsync/__init__.py
glennmatthews: In my view this would be a coding/usage error if you're calling it with a nonexistent attribute, so I don't think we need to capture the AttributeError.

glennmatthews (dismissed):
Great stuff!

FragmentedPacket:
@dgarros Not sure if you want to review it before it gets merged in.

dgarros (changes requested):
LGTM, great work @FragmentedPacket and thanks helping with this one

One last change for me before merging
please. can you add a README.md in the new example directory (04-XX) and add a link to this readme file in the documentation > `diffsync/docs/source/examples/index.rst`

FragmentedPacket:
@dgarros I completely forgot to even write any docs for this so thanks for catching that.

FragmentedPacket:
Does this also address #73? Just want to make sure once this is merged, we can close it.

dgarros (approved):
Looks great, thanks @FragmentedPacket

dgarros:
> Does this also address #73? Just want to make sure once this is merged, we can close it.

I think it does

dgarros:
I'll merge since Glenn already approved it

Issue #74, opened 2021-10-28
Support current structlog major version.


grelleum:
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
Currently the `pyproject.toml` file restricts `structlog` the major version to 20.
`structlog = "^20.1.0"`
The current major version is 21.

### Use Case
For my project, there is another required package that has pinned a minimum structlog version of 21.0.0.  When adding diffsync as a requirement to my project, pip will install a much older version of the other required package where the minimum structlog version was 20.  This causes an issue with needed features not being available.

dgarros:
Fixed by #75

PR #75, opened 2021-10-28
Allow installation of current 'structlog' major version 21.x

grelleum:
This PR allows using `structlog` major version 21 to be installed with `diffsync`.
This fixes issue: https://github.com/networktocode/diffsync/issues/74

Three changes are made:
1. Pinned the `pylint` version to 2.6.x so that `invoke tests` will complete successfully after running `poetry lock`.
2. Modified `structlog` versioning to allow 20.x.x or 21.x.x.  Ran complete tests to verify compatibility.
3. Updated `poetry.lock` file to include current `structlog` version.

### DIFF
--- a/diffsync/logging.py
+++ b/diffsync/logging.py
@@ -58,7 +58,9 @@ def enable_console_logging(verbosity=0):
 
 
 def _structlog_exception_formatter_required():
-    """Return True if structlog exception formatter should be loaded
+    """Determine if structlog exception formatter is needed.
+
+    Return True if structlog exception formatter should be loaded
     into structlog processors.
 
     Structlog version 21.2.0 or higher will generate a warning
@@ -68,7 +70,6 @@ def _structlog_exception_formatter_required():
     This code snippet will determine if we need to add 'format_exc_info'
     to the processors.
     """
-
     structlog_float_version = float(".".join(structlog.__version__.split(".")[:2]))
     if structlog_float_version < 21.2:
         return True
--- a/diffsync/logging.py
+++ b/diffsync/logging.py
@@ -15,6 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from packaging import version
 import importlib
 import logging
 
@@ -71,8 +72,8 @@ def _structlog_exception_formatter_required():
     This code snippet will determine if we need to add 'format_exc_info'
     to the processors.
     """
-    structlog_float_version = float(".".join(structlog.__version__.split(".")[:2]))
-    if structlog_float_version < 21.2:
+
+    if version.parse(structlog.__version__) < version.Version("21.2.0"):
         return True
 
     # Determine if module is available for import, without importing it.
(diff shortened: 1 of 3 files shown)


grelleum:
Thanks @dgarros - I will look further into this as see how these changes affects DiffSync.

diffsync/logging.py
```diff
@@ -35,18 +35,52 @@ def enable_console_logging(verbosity=0):
...
+    if either rich or better_exceptions packages are available to import
+    when the 'format_exc_info' processor is used.
+
+    This code snippet will determine if we need to add 'format_exc_info'
+    to the processors.
+    """
+
+    structlog_float_version = float(".".join(structlog.__version__.split(".")[:2]))
+    if structlog_float_version < 21.2:
+        return True
```
glennmatthews, line 74: This looks pretty gross. Can we use `packaging.Version` objects here instead? Something like:

```python
from packaging import version

if version.parse(structlog.__version__) < version.Version("21.2.0"):
```

grelleum: Great suggestion - I've implemented the change.

diffsync/logging.py
```diff
@@ -35,18 +38,45 @@ def enable_console_logging(verbosity=0):
...
     structlog.configure(
-        processors=[
-            structlog.stdlib.filter_by_level,  # <-- added
-            structlog.stdlib.add_logger_name,
-            structlog.stdlib.add_log_level,
-            structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M.%S"),
-            structlog.processors.StackInfoRenderer(),
-            structlog.processors.format_exc_info,
-            structlog.dev.ConsoleRenderer(),
-        ],
```
glennmatthews, line 47: Don't we need a `processors=processors,` line added here?

grelleum:
The structlog change in 21.2.0 and above changes behavior with regards to the use of `format_exc_info` as a log processor.
This effects exception logging when using the `logging.enable_console_logging` function.

When using 21.2.0 or newer and either `better_exceptions` or `rich` in available in the Python environment, then exception logging will generate a warning if `structlog.processors.format_exc_info` is installed as a processor.  If we remove the `format_exc_info` processor then we lose exception traceback formatting, when the above is not the case.

The code added here determines if the `structlog.processors.format_exc_info` is required in the processor list, and appends it to the list if required.

diffsync/logging.py
grelleum: Indeed, it got lost somewhere in the experimentation phase.  Added this back in.

grelleum:
For compatibility with Python 3.9, I changed:
```
import importlib
...
    rich = importlib.util.find_spec("rich")
    better_exceptions = importlib.util.find_spec("better_exceptions")
```
to
```
from importlib.util import find_spec
... 
    rich = find_spec("rich")
    better_exceptions = find_spec("better_exceptions")
```

dgarros (approved):
LGTM thank you @grelleum for the PR
@glennmatthews do you think we are good to merge ?

glennmatthews (approved):
:shipit:

PR #80, opened 2021-11-14
Extend sync_to|from to accept a Diff

### LINKED ISSUE #76: Extend `sync_from` and `sync_to` to accept an existing diff
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
Currently doing a `diff_to/from` followed by a `sync_to/from`, result in calculating the diff twice because `sync_to/from` are calculating a new diff automatically.
The proposal is to extend `sync_from` and `sync_to` to accept an existing diff

```python
    self.log_info(message="Loading current data from Data Source...")
    diffsync1 = DataSourceDiffSync(job=self, sync=self.sync)
    diffsync1.load()

    self.log_info(message="Loading current data from Nautobot...")
    diffsync2 = NautobotDiffSync(job=self, sync=self.sync)
    diffsync2.load()

    diffsync_flags = DiffSyncFlags.CONTINUE_ON_FAILURE

    self.log_info(message="Calculating diffs...")
    diff = diffsync1.diff_to(diffsync_1, flags=diffsync_flags)

    if not self.kwargs["dry_run"]:
        self.log_info(message="Syncing from Data Source to Nautobot...")
        diffsync1.sync_to(diffsync2, flags=diffsync_flags, diff=diff). <<<<<<<<<<<
        self.log_info(message="Sync complete")
```

### Use Case
Performance improvement, there is no need to calculate the diff twice on the same dataset

FragmentedPacket:
Fixes #76 

This allows a user to pass in an existing `Diff` object into `sync_to|from` to prevent another diff from being calculated.

- Updated examples, docs, and tests.
- Remove print statement within tests

### DIFF
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -461,7 +461,8 @@ class DiffSync:
         diff_class: Type[Diff] = Diff,
         flags: DiffSyncFlags = DiffSyncFlags.NONE,
         callback: Optional[Callable[[Text, int, int], None]] = None,
-    ):
+        diff: Diff = None,
+    ):  # pylint: disable=too-many-arguments:
         """Synchronize data from the given source DiffSync object into the current DiffSync object.
 
         Args:
@@ -470,8 +471,10 @@ class DiffSync:
             flags (DiffSyncFlags): Flags influencing the behavior of this sync.
             callback (function): Function with parameters (stage, current, total), to be called at intervals as the
                 calculation of the diff and subsequent sync proceed.
+            diff (Diff): An existing diff to be used rather than generating a completely new diff.
         """
-        diff = self.diff_from(source, diff_class=diff_class, flags=flags, callback=callback)
+        if not diff:
+            diff = self.diff_from(source, diff_class=diff_class, flags=flags, callback=callback)
         syncer = DiffSyncSyncer(diff=diff, src_diffsync=source, dst_diffsync=self, flags=flags, callback=callback)
         result = syncer.perform_sync()
         if result:
@@ -483,7 +486,8 @@ class DiffSync:
         diff_class: Type[Diff] = Diff,
         flags: DiffSyncFlags = DiffSyncFlags.NONE,
         callback: Optional[Callable[[Text, int, int], None]] = None,
-    ):
+        diff: Diff = None,
+    ):  # pylint: disable=too-many-arguments
         """Synchronize data from the current DiffSync object into the given target DiffSync object.
 
         Args:
@@ -492,8 +496,9 @@ class DiffSync:
             flags (DiffSyncFlags): Flags influencing the behavior of this sync.
             callback (function): Function with parameters (stage, current, total), to be called at intervals as the
                 calculation of the diff and subsequent sync proceed.
+            diff (Diff): An existing diff that will be used when determining what needs to be synced.
         """
-        target.sync_from(self, diff_class=diff_class, flags=flags, callback=callback)
+        target.sync_from(self, diff_class=diff_class, flags=flags, callback=callback, diff=diff)
 
     def sync_complete(
         self,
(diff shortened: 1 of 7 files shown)


examples/01-multiple-data-sources/README.md
```diff
@@ -57,6 +57,8 @@ Synchronize A and B (update B with the contents of A):
 ```python
 a.sync_to(b)
 print(a.diff_to(b).str())
+# Alternatively you can pass in the diff object from above to prevent another diff calculation
+a.sync_to(b, diff_a_b)
```
glennmatthews, line 61:
```suggestion
# a.sync_to(b, diff=diff_a_b)
```

examples/03-remote-system/main.py
```diff
@@ -41,7 +41,7 @@ def main():
 
     if args.sync:
         print("Updating the list of countries in Nautobot ...")
-        nautobot.sync_from(local, flags=flags, diff_class=AlphabeticalOrderDiff)
+        nautobot.sync_from(local, flags=flags, diff_class=AlphabeticalOrderDiff, diff=diff)
```
glennmatthews, line 44: `diff` will be an undefined variable if you didn't specify `args.diff`, won't it?

tests/unit/test_diffsync.py
```diff
@@ -468,6 +468,26 @@ def callback(stage, current, total):
     assert last_value == {"current": expected, "total": expected}
 
 
+def test_diffsync_sync_to_w_diff(backend_a, backend_b):
+    diff = backend_b.diff_to(backend_a)
+    assert diff.has_diffs()
+    # Mock diff_from to make sure it's not called when passing in an existing diff
+    backend_b.diff_from = mock.Mock()
```
glennmatthews, line 475: For extra certainty, maybe mock both `diff_from` and `diff_to` on both `backend_a` and `backend_b` and assert that none of the four get called?

tests/unit/test_diffsync.py
```diff
@@ -468,6 +468,26 @@ def callback(stage, current, total):
...
+    # Perform full sync
+    backend_b.sync_to(backend_a, diff=diff)
+    assert not backend_b.diff_from.called
```
glennmatthews, line 478: Might also be good to assert that the sync actually updates `backend_a` according to the provided diff?

diffsync/__init__.py
```diff
@@ -461,7 +461,8 @@ def sync_from(
         diff_class: Type[Diff] = Diff,
         flags: DiffSyncFlags = DiffSyncFlags.NONE,
         callback: Optional[Callable[[Text, int, int], None]] = None,
-    ):
+        diff: Diff = None,
```
glennmatthews, line 464:
```suggestion
        diff: Optional[Diff] = None,
```

diffsync/__init__.py
```diff
@@ -483,7 +486,8 @@ def sync_to(
...
```
glennmatthews, line 489:
```suggestion
        diff: Optional[Diff] = None,
```

diffsync/__init__.py
```diff
@@ -470,8 +471,10 @@ def sync_from(
             flags (DiffSyncFlags): Flags influencing the behavior of this sync.
             callback (function): Function with parameters (stage, current, total), to be called at intervals as the
                 calculation of the diff and subsequent sync proceed.
+            diff (Diff): An existing diff to be used rather than generating a completely new diff.
         """
-        diff = self.diff_from(source, diff_class=diff_class, flags=flags, callback=callback)
+        if not diff:
+            diff = self.diff_from(source, diff_class=diff_class, flags=flags, callback=callback)
```
glennmatthews, line 477: If both `diff` and `diff_class` parameters are non-None, should we assert that `isinstance(diff, diff_class)`? Or should we disallow specifying `diff_class` at all if `diff` was specified?

FragmentedPacket: Good catch! I think it would make sense to compare and raise an error if they don't match. The reason being is that we want to make sure the user gets their intended synchronization by making sure the diff they passed in is subclassed by the diff_class that is passed in or that is default.

tests/unit/test_diffsync.py
FragmentedPacket: I had to create this within a new test since we're mocking within this test, but it's there now to do post diff and assert it doesn't have any diffs.

glennmatthews (approved):
Looks good!

PR #81, opened 2021-11-16
Example 5: PeeringDB to Nautobot example

chadell:
Contributing with this example: https://github.com/chadell/playground-diffsync

### DIFF
--- a/examples/05-nautobot-peeringdb/README.md
+++ b/examples/05-nautobot-peeringdb/README.md
@@ -1,0 +1,65 @@
+# playground-diffsync
+
+## Description
+
+This repository is just a set of examples around [diffsync](https://github.com/networktocode/diffsync) in order to understand how it works and how you could use it.
+
+## Install dependencies
+
+```bash
+python3 -m venv .venv
+source .venv/bin/activate
+pip3 install -r requirements.txt
+```
+
+## Context
+
+The goal of this example is to synchronize some data from [PeeringDB](https://www.peeringdb.com/), that as the name suggests is a DB where peering entities define their facilities and presence to facilitate peering, towards [Nautobot Demo](https://demo.nautobot.com/) that is a always on demo service for [Nautobot](https://nautobot.readthedocs.io/), an open source Source of Truth.
+
+In Peering DB there is a model that defines a `Facility` and you can get information about the actual data center and the city where it is placed. In Nautobot, this information could be mapped to the `Region` and `Site` models, where `Region` can define hierarchy. For instance, Barcelona is in Spain and Spain is in Europe, and all of them are `Regions`. And, finally, the actual datacenter will refer to the `Region` where it is placed.
+
+Because of the nature of the demo, we will focus on syncing from PeeringDB to Nautobot (we can assume that PeeringDB is the authoritative System of Record) and we will skip the `delete` part of the `diffsync` library.
+
+## Show me the code
+
+We have 3 files:
+
+- `models.py`: defines the reference models that we will use: `RegionMode` and `SiteModel`
+- `adapter_peeringdb.py`: defines the PeeringDB adapter to translate via `load()` the data from PeeringDB into the reference models commented above. Notice that we don't define CRUD methods because we will sync from it (no to it)
+- `adapter_nautobot.py`: deifnes the Nautobot adapter with the `load()` and the CRUD methods
+
+```python
+from IPython import embed
+embed(colors="neutral")
+
+# Import Adapaters
+from adapter_nautobot import NautobotRemote
+from adapter_peeringdb import PeeringDB
+
+# Initialize PeeringDB adapter, using CATNIX id for demonstration
+peeringdb = PeeringDB(ix_id=62)
+
+# Initialize Nautobot adapter, pointing to the demo instance (it's also the default settings)
+nautobot = NautobotRemote(
+    url="https://demo.nautobot.com",
+    token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+)
+
+# Load PeeringDB info into the adapter
+peeringdb.load()
+
+# We can check the data that has been imported, some as `site` and some as `region` (with the parent relationships)
+peeringdb.dict()
+
+# Load Nautobot info into the adapter
+nautobot.load()
+
+# Let's diffsync do it's magic
+diff = nautobot.diff_from(peeringdb)
+
+# Quick summary of the expected changes (remember that delete ones are dry-run)
+diff.summary()
+
+# Execute the synchronization
+nautobot.sync_from(peeringdb)
+```
--- a/examples/05-nautobot-peeringdb/README.md
+++ b/examples/05-nautobot-peeringdb/README.md
@@ -24,13 +24,15 @@ source .venv/bin/activate
 pip3 install -r requirements.txt
 ```
 
-## Run it interactivelly
+## Run it interactively
 
 ```python
 from IPython import embed
 embed(colors="neutral")
 
-# Import Adapaters
+# Import Adapters
+from diffsync.enum import DiffSyncFlags
+
 from adapter_nautobot import NautobotRemote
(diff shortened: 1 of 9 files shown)


examples/05-nautobot-peeringdb/adapter_nautobot.py
```diff
@@ -0,0 +1,201 @@
...
+            attrs (dict): Initial values for this model's _attributes
+        """
+        data = {
+            "name": ids["name"],
+            "slug": attrs["slug"],
+        }
+        if attrs["description"]:
+            data["description"] = attrs["description"]
+        if attrs["parent_name"]:
+            data["parent"] = str(diffsync.get(diffsync.region, attrs["parent_name"]).pk)
```
glennmatthews, line 30: Could we also do:

```python
data["parent"] = {"name": attrs["parent_name"]}
```

as in the `update` case below? Or is there a reason for the different approach between these two methods?

examples/05-nautobot-peeringdb/adapter_nautobot.py
```diff
@@ -0,0 +1,201 @@
...
+        self.headers = {
+            "Accept": "application/json",
+            "Authorization": f"Token {self.token}",
+        }
+
+    def load(self):
+        """Load Region and Site data from the remote Nautobot instance."""
+        # To keep the example simple, we disable REST API pagination of results.
+        # In a real job, especially when expecting a large number of results,
+        # we'd want to handle pagination, or use something like PyNautobot.
```
glennmatthews, line 147: Aren't you actually handling pagination with your `region_data["next"]` logic below?

examples/05-nautobot-peeringdb/adapter_nautobot.py
chadell: I only tried the `create` for this model... so the `update` was not updated accordingly, will do.

examples/05-nautobot-peeringdb/adapter_nautobot.py
chadell: yes, not updated properly. you can guess where this copy&paste is coming from

examples/05-nautobot-peeringdb/adapter_peeringdb.py
```diff
@@ -0,0 +1,55 @@
...
+
+        for fac in ix_data["data"][0]["fac_set"]:
+            # PeeringDB has no Region entity, so we must avoid duplicates
+            try:
+                self.get(self.region, fac["city"])
+            except ObjectNotFound:
+                region = self.region(
+                    name=fac["city"],
+                    slug=slugify(fac["city"]),
+                    parent_name=pycountry.countries.get(alpha_2=fac["country"]).name,
```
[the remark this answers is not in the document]
chadell: yes, true. For the demo I didn't need that because the parent was already available :) 
but gonna improve it

examples/05-nautobot-peeringdb/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Let's diffsync do it's magic
+diff = nautobot.diff_from(peeringdb)
+
+# Quick summary of the expected changes (remember that delete ones are dry-run)
+diff.summary()
+
+# Execute the synchronization
+nautobot.sync_from(peeringdb)
+```
```
glennmatthews, line 65: For completeness as an example, we should include a `main.py` that basically does exactly the above code, and add a test case in `tests/unit/test_examples.py` that runs the main module and verifies that it succeeds.

glennmatthews (dismissed):

chadell:
sorry @glennmatthews , I requested review but I forgot that I had pending changes (`main.py`) to push...
Will ping you when ready

glennmatthews:
Good thought. Maybe make the `main.py` (which it looks like you may not have committed to Git?) only run up to the `diff.summary()` and not perform the final `sync_from()` step?

examples/05-nautobot-peeringdb/README.md
```diff
@@ -0,0 +1,63 @@
...
+nautobot.load()
+
+# Let's diffsync do it's magic
+diff = nautobot.diff_from(peeringdb)
+
+# Quick summary of the expected changes (remember that delete ones are dry-run)
+diff.summary()
+
+# Execute the synchronization
+nautobot.sync_from(peeringdb)
```
glennmatthews, line 62: Since we're not implementing `delete()`, it might be good as an example to change this to explicitly skip deletion?

```suggestion
from diffsync.enum import DiffSyncFlags
nautobot.sync_from(peeringdb, flags=DiffSyncFlags.SKIP_UNMATCHED_DST)
```

examples/05-nautobot-peeringdb/adapter_nautobot.py
```diff
@@ -0,0 +1,201 @@
...
+        for region_entry in regions:
+            region = self.region(
+                name=region_entry["name"],
+                slug=region_entry["slug"],
+                description=region_entry["description"] or None,
+                parent_name=region_entry["parent"]["name"] if region_entry["parent"] else None,
+                pk=region_entry["id"],
+            )
+            self.add(region)
+            # self.job.log_debug(message=f"Loaded {region} from remote Nautobot instance")
```
glennmatthews, line 163: To avoid reader confusion, would you please remove this and the similar `job.log_debug` comment below?

examples/05-nautobot-peeringdb/adapter_peeringdb.py
```diff
@@ -0,0 +1,66 @@
...
+        """Load data via from PeeringDB."""
+        ix_data = requests.get(f"{PEERINGDB_URL}/api/ix/{self.ix_id}").json()
+
+        for fac in ix_data["data"][0]["fac_set"]:
+            # PeeringDB has no Region entity, so we must avoid duplicates
+            try:
+                self.get(self.region, fac["city"])
+            except ObjectNotFound:
+                # Adding the parent region if necessary
+                parent_name = pycountry.countries.get(alpha_2=fac["country"]).name
```
glennmatthews, line 39: Maybe elaborate on the comment here as this isn't immediately obvious to users who aren't familiar with `pycountry`, something like:

```suggestion
                # Use pycountry to translate the country code (like "DE") to a country name (like "Germany")
                parent_name = pycountry.countries.get(alpha_2=fac["country"]).name
                # Add the country as a parent region if not already added
```

chadell:
> # Use pycountry to translate the country code (like "DE") to a country name (like "Germany")
>                 parent_name = pycountry.countries.get(alpha_2=fac["country"]).name
>                 # Add the country as a parent region if not already added

Yes, I believe that in this case is wiser to skip the test as being dependent on external systems. I added the `main.py` as reference

glennmatthews (dismissed):

dgarros (dismissed):
Looks good to me overall, 
it's not a blocker but I think if we should leverage pynautobot whenever possible, it makes the  code cleaner.

chadell:
> Looks good to me overall, it's not a blocker but I think if we should leverage pynautobot whenever possible, it makes the code cleaner.

I did it this way to avoid the extra dependency

dgarros:
> I did it this way to avoid the extra dependency

That's a good reminder that we should be consistent here
for the other examples, we didn't add the dependencies to the `pyproject.toml`, instead we have [a requirements.txt file in the example directory](https://github.com/networktocode/diffsync/blob/main/examples/03-remote-system/requirements.txt)

I'm good either way but we should be consistent

@glennmatthews what is your preference ?

dgarros (commented):
@chadell please can you remove the changes from the pyproject.toml and add a requirements.txt file in your example directory to align with the other examples

chadell:
> @chadell please can you remove the changes from the pyproject.toml and add a requirements.txt file in your example directory to align with the other examples

The `requirements.txt` file is already present. the changes in `pyproject.toml` should not be there, will remove.

dgarros (approved):
Thank you @chadell

PR #82, opened 2021-12-14
Update lock file with latest version

dgarros:
This PR only update the lock file that we are using for testing so it shouldn't affect the end users of the library
In the process I had to change some code to align with the latest version of pylint

### DIFF
--- a/examples/03-remote-system/local_adapter.py
+++ b/examples/03-remote-system/local_adapter.py
@@ -29,7 +29,7 @@ class LocalAdapter(DiffSync):
 
     def load(self, filename=COUNTRIES_FILE):  # pylint: disable=arguments-differ
         """Load all regions and countries from a local JSON file."""
-        with open(filename, "r") as data_file:  # pylint: disable=unspecified-encoding
+        with open(filename, "r", encoding="UTF-8") as data_file:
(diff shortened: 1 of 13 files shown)


examples/03-remote-system/local_adapter.py
```diff
@@ -29,7 +29,7 @@ class LocalAdapter(DiffSync):
 
     def load(self, filename=COUNTRIES_FILE):  # pylint: disable=arguments-differ
         """Load all regions and countries from a local JSON file."""
-        with open(filename, "r") as data_file:
+        with open(filename, "r") as data_file:  # pylint: disable=unspecified-encoding
```
glennmatthews, line 32: Can we just specify utf-8 as the encoding instead?

tasks.py
```diff
@@ -27,7 +27,7 @@
 
 def project_ver():
     """Find version from pyproject.toml to use for docker image tagging."""
-    with open("pyproject.toml") as file:
+    with open("pyproject.toml") as file:  # pylint: disable=unspecified-encoding
```
glennmatthews, line 30: utf-8 here as well?

glennmatthews (dismissed):

PR #83, opened 2021-12-15
Add python 3.10 to CI pipeline

dgarros:


glennmatthews (approved):
Seems fine - is there a particular reason for this addition?

dgarros:
Just house keeping to ensure that we are not introducing something that wouldn't work with 3.10
also, before #82 it wasn't passing

Issue #49, opened 2021-04-08
Option to skip deletion of records on a case-by-case basis


glennmatthews:
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality

Add a model flag that can be used to control whether an unmatched model class or instance will trigger deletion (and/or creation) of records when a sync operation is run.

### Use Case

The existing global IGNORE_UNMATCHED_DST flag is not sufficiently granular as it applies to all records and all models. In some cases that may be adequate, but in others there needs to be per-model or even per-record control over this behavior -- for example, an application may not wish to delete unmatched Device records (perhaps a device is temporarily offline and hence not included in the source data), but may still wish to delete unmatched Interface records (as they reflect incorrect information about existing devices).

dgarros:
@glennmatthews I think this has been fixed by #87

PR #87, opened 2021-12-16
Add new model flags

### LINKED ISSUE #85: Add model flags to control which crud method should be executed on sync
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
It would be very useful to have a few more Model flags to control which CRUD methods (create/update/delete) would be called during a sync()

I would like to propose 
- `CRUD_NO_UPDATE`: Do not call update() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff.
- `CRUD_NO_DELETE`: Do not call delete() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff.
- `CRUD_NO_UPDATE_DELETE` = `CRUD_NO_UPDATE` | `CRUD_NO_DELETE`

I wish we could support CRUD_NO_CREATE but I don't think this is possible right now because we can't pass context to this method since the model doesn't exist yet.

### Use Case
The main use case for me would be to `protect` some object to be `READ_ONLY` but still show up in the diff while other objects of the same type would remain `READ_WRITE`.
As an example
When using the network-importer, once we have done the initial import of the SOT and the data has been cleaned up, it would be useful to `protect` some objects from being updated/deleted in the SOT but it's still interesting to have these objects show up in the diff.
Today for this use case we are using the flag `IGNORE` but the object is completely ignored and doesn't show up in the diff at all.

dgarros:
Fixes #85 & #49

This PR adds 6 new model flags
- `SKIP_UNMATCHED_SRC` : Ignore the model if it only exists in the source/"from" DiffSync when determining diffs and syncing.
- `SKIP_UNMATCHED_DST` : Ignore the model if it only exists in the target/"to" DiffSync when determining diffs and syncing.
- `SKIP_UNMATCHED_BOTH`: `SKIP_UNMATCHED_SRC` | `SKIP_UNMATCHED_DST`
- `CRUD_NO_UPDATE`: Do not call update() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff.
- `CRUD_NO_DELETE`: Do not call delete() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff.
- `CRUD_NO_UPDATE_DELETE` = `CRUD_NO_UPDATE` | `CRUD_NO_DELETE`
 
The new model flags `SKIP_UNMATCHED_*` provide the same behavior as the global flags with the same name.

The `CRUD_*` model flags provide a way to disable some CURD method selectively. 
Since, in the current design the `create`/`update`/`delete` method must execute the base methods with `super()` I had to move the logic in the base methods into a new method `create|update|delete_base()` in order to not execute the user code but still keep the same behavior internally. Not sure if we should recommend the user to call new base methods now or if it's best to keep using `super()` 
 
Also one of the unit test file got too long (1000 lines) so I had to split it into 2 files

### TODO
- [x] code
- [x] unit tests 
- [x] update documentation

### DIFF
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -176,10 +176,28 @@ class DiffSyncModel(BaseModel):
         self._status_message = message
 
     @classmethod
+    def create_base(cls, diffsync: "DiffSync", ids: Mapping, attrs: Mapping) -> Optional["DiffSyncModel"]:
+        """Instantiate this class, along with any platform-specific data creation.
+
+        This method is not meant to be subclassed, users should redefine create() instead.
+
+        Args:
+            diffsync: The master data store for other DiffSyncModel instances that we might need to reference
+            ids: Dictionary of unique-identifiers needed to create the new object
+            attrs: Dictionary of additional attributes to set on the new object
+
+        Returns:
+            DiffSyncModel: instance of this class.
+        """
+        model = cls(**ids, diffsync=diffsync, **attrs)
+        model.set_status(DiffSyncStatus.SUCCESS, "Created successfully")
+        return model
+
+    @classmethod
     def create(cls, diffsync: "DiffSync", ids: Mapping, attrs: Mapping) -> Optional["DiffSyncModel"]:
         """Instantiate this class, along with any platform-specific data creation.
 
-        Subclasses must call `super().create()`; they may wish to then override the default status information
+        Subclasses must call `super().create()` or `self.create_base()`; they may wish to then override the default status information
         by calling `set_status()` to provide more context (such as details of any interactions with underlying systems).
 
         Args:
@@ -194,14 +212,30 @@ class DiffSyncModel(BaseModel):
         Raises:
             ObjectNotCreated: if an error occurred.
         """
-        model = cls(**ids, diffsync=diffsync, **attrs)
-        model.set_status(DiffSyncStatus.SUCCESS, "Created successfully")
-        return model
+        return cls.create_base(diffsync=diffsync, ids=ids, attrs=attrs)
+
+    def update_base(self, attrs: Mapping) -> Optional["DiffSyncModel"]:
+        """Base Update method to update the attributes of this instance, along with any platform-specific data updates.
+
+        This method is not meant to be subclassed, users should redefine update() instead.
+
+        Args:
+            attrs: Dictionary of attributes to update on the object
+
+        Returns:
+            DiffSyncModel: this instance.
+        """
+        for attr, value in attrs.items():
+            # TODO: enforce that only attrs in self._attributes can be updated in this way?
+            setattr(self, attr, value)
+
+        self.set_status(DiffSyncStatus.SUCCESS, "Updated successfully")
+        return self
 
     def update(self, attrs: Mapping) -> Optional["DiffSyncModel"]:
         """Update the attributes of this instance, along with any platform-specific data updates.
 
-        Subclasses must call `super().update()`; they may wish to then override the default status information
+        Subclasses must call `super().update()` or `self.update_base()`; they may wish to then override the default status information
         by calling `set_status()` to provide more context (such as details of any interactions with underlying systems).
 
         Args:
@@ -214,17 +248,23 @@ class DiffSyncModel(BaseModel):
         Raises:
             ObjectNotUpdated: if an error occurred.
         """
-        for attr, value in attrs.items():
-            # TODO: enforce that only attrs in self._attributes can be updated in this way?
-            setattr(self, attr, value)
+        return self.update_base(attrs=attrs)
 
-        self.set_status(DiffSyncStatus.SUCCESS, "Updated successfully")
+    def delete_base(self) -> Optional["DiffSyncModel"]:
+        """Base delete method Delete any platform-specific data corresponding to this instance.
+
+        This method is not meant to be subclassed, users should redefine delete() instead.
+
+        Returns:
+            DiffSyncModel: this instance.
+        """
+        self.set_status(DiffSyncStatus.SUCCESS, "Deleted successfully")
         return self
 
     def delete(self) -> Optional["DiffSyncModel"]:
         """Delete any platform-specific data corresponding to this instance.
 
-        Subclasses must call `super().delete()`; they may wish to then override the default status information
+        Subclasses must call `super().delete()` or `self.delete_base()`; they may wish to then override the default status information
         by calling `set_status()` to provide more context (such as details of any interactions with underlying systems).
 
         Returns:
@@ -234,8 +274,7 @@ class DiffSyncModel(BaseModel):
         Raises:
             ObjectNotDeleted: if an error occurred.
         """
-        self.set_status(DiffSyncStatus.SUCCESS, "Deleted successfully")
-        return self
+        return self.delete_base()
 
     @classmethod
     def get_type(cls) -> Text:
--- a/diffsync/enum.py
+++ b/diffsync/enum.py
@@ -35,6 +35,20 @@ class DiffSyncModelFlags(enum.Flag):
     Can be used for the case where deletion of a model results in the automatic deletion of all its children.
     """
 
+    SKIP_UNMATCHED_SRC = 0b100
+    """Ignore the model if it only exists in the source/"from" DiffSync when determining diffs and syncing.
+
+    If this flag is set, no new model will be created in the target/"to" DiffSync.
+    """
+
+    SKIP_UNMATCHED_DST = 0b1000
+    """Ignore the model if it only exists in the target/"to" DiffSync when determining diffs and syncing.
+
+    If this flag is set, the model will not be deleted from the target/"to" DiffSync.
+    """
+
+    SKIP_UNMATCHED_BOTH = SKIP_UNMATCHED_SRC | SKIP_UNMATCHED_DST
+
 
 class DiffSyncFlags(enum.Flag):
     """Flags that can be passed to a sync_* or diff_* call to affect its behavior."""
--- a/diffsync/helpers.py
+++ b/diffsync/helpers.py
@@ -408,7 +408,7 @@ class DiffSyncSyncer:  # pylint: disable=too-many-instance-attributes
             return (False, dst_model)
 
         try:
-            self.logger.debug(f"Attempting model {self.action}")
+            self.logger.debug(f"Attempting model {self.action.value}")
             if self.action == DiffSyncActions.CREATE:
                 if dst_model is not None:
                     raise ObjectNotCreated(f"Failed to create {self.model_class.get_type()} {ids} - it already exists!")
--- a/diffsync/helpers.py
+++ b/diffsync/helpers.py
@@ -155,7 +155,7 @@ class DiffSyncDiffer:  # pylint: disable=too-many-instance-attributes
                 if src_obj.get_identifiers() != dst_obj.get_identifiers():
                     raise ValueError(f"Keys mismatch: {src_obj.get_identifiers()} vs {dst_obj.get_identifiers()}")
 
-    def diff_object_pair(
+    def diff_object_pair(  # pylint: disable=too-many-return-statements
         self, src_obj: Optional["DiffSyncModel"], dst_obj: Optional["DiffSyncModel"]
     ) -> Optional[DiffElement]:
         """Diff the two provided DiffSyncModel objects and return a DiffElement or None.
@@ -180,11 +180,19 @@ class DiffSyncDiffer:  # pylint: disable=too-many-instance-attributes
 
         log = self.logger.bind(model=model, unique_id=unique_id)
         if self.flags & DiffSyncFlags.SKIP_UNMATCHED_SRC and not dst_obj:
-            log.debug("Skipping unmatched source object")
+            log.debug("Skipping due to SKIP_UNMATCHED_SRC flag on source adapter")
             self.incr_models_processed()
             return None
         if self.flags & DiffSyncFlags.SKIP_UNMATCHED_DST and not src_obj:
-            log.debug("Skipping unmatched dest object")
+            log.debug("Skipping due to SKIP_UNMATCHED_DST flag on source adapter")
+            self.incr_models_processed()
+            return None
+        if src_obj and not dst_obj and src_obj.model_flags & DiffSyncModelFlags.SKIP_UNMATCHED_SRC:
+            log.debug("Skipping due to SKIP_UNMATCHED_SRC flag on model")
+            self.incr_models_processed()
+            return None
+        if dst_obj and not src_obj and dst_obj.model_flags & DiffSyncModelFlags.SKIP_UNMATCHED_DST:
+            log.debug("Skipping due to SKIP_UNMATCHED_DST flag on model")
             self.incr_models_processed()
             return None
         if src_obj and src_obj.model_flags & DiffSyncModelFlags.IGNORE:
@@ -284,6 +292,7 @@ class DiffSyncSyncer:  # pylint: disable=too-many-instance-attributes
     ):
         """Create a DiffSyncSyncer instance, ready to call `perform_sync()` against."""
         self.diff = diff
+        self.src_diffsync = src_diffsync
         self.dst_diffsync = dst_diffsync
         self.flags = flags
         self.callback = callback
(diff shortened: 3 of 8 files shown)


glennmatthews:
The `_base` methods feel clunky to me. I think in general we should (possibly as a breaking "2.0" change) refine the create/update/delete methods to accept `**kwargs` so that we can add additional passthrough parameters here and the future without it being a breaking change again.

diffsync/helpers.py
```diff
@@ -398,11 +406,17 @@ def sync_model(
             elif self.action == "update":
                 if model is None:
                     raise ObjectNotUpdated(f"Failed to update {self.model_class.get_type()} {ids} - not found!")
-                model = model.update(attrs=attrs)
+                if model.model_flags & DiffSyncModelFlags.CRUD_NO_UPDATE:
+                    model = model.update_base(attrs=attrs)
```
glennmatthews, line 410: `update_base` still updates the DiffSyncModel instance with the new `attrs`? To me that doesn't align with the concept of `CRUD_NO_UPDATE` as a label. I would expect that it would make no changes to the DiffSyncModel at all... consider also the case of repeated syncing/diffing with these flags in play:

- `diff()` - shows model delta in the diffs, even the models that set CRUD_NO_*
- `sync()` - calls `_base` methods on all models, resulting in DiffSyncModel instances being created/updated/deleted
- `diff()` - now shows no diffs at all, even on the CRUD_NO_* models (because they were modified by `sync()`)

That seems wrong to me.

diffsync/helpers.py
```diff
@@ -180,13 +180,21 @@ def diff_object_pair(
 
         log = self.logger.bind(model=model, unique_id=unique_id)
         if self.flags & DiffSyncFlags.SKIP_UNMATCHED_SRC and not dst_obj:
-            log.debug("Skipping unmatched source object")
+            log.debug("Skipping due to SKIP_UNMATCHED_SRC flag on source adapter")
             self.incr_models_processed()
             return None
         if self.flags & DiffSyncFlags.SKIP_UNMATCHED_DST and not src_obj:
             log.debug("Skipping unmatched dest object")
```
glennmatthews, line 187: Updating this message as well?

diffsync/helpers.py
dgarros: Good point, I've updated the logic to not touch the local object as well

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,196 @@
...
+    # Validate that there are 2 extras objects out of the box
+    diff = backend_a.diff_from(backend_a_with_extra_models)
+    assert diff.summary() == {"create": 2, "update": 0, "delete": 0, "no-change": 23}
+
+    # Check that only 1 object is affected by the flag
+    backend_a_with_extra_models.get(
+        backend_a_with_extra_models.site, "lax"
+    ).model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_SRC
+    diff = backend_a.diff_from(backend_a_with_extra_models)
+    assert diff.summary() == {"create": 1, "update": 0, "delete": 0, "no-change": 23}
```
glennmatthews, line 34: Seems weird to me that setting the flag makes this model "disappear" from the summary counts. Offhand I would expect that the diff summary should still show a "create" for that model - but if that's not the case, should it instead show as a "no-change" or something?

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,196 @@
...
+
+    backend_a.sync_from(backend_a_with_extra_models)
+
+    # New objects should not have been created
+    with pytest.raises(ObjectNotFound):
+        backend_a.get(backend_a.site, "lax")
+    with pytest.raises(ObjectNotFound):
+        backend_a.get(backend_a.device, "nyc-spine3")
+    assert "nyc-spine3" not in backend_a.get(backend_a.site, "nyc").devices
+
```
glennmatthews, line 60: Post-sync, what do we expect a diff between the two instances to show?

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,196 @@
...
+    # Check that only 1 object is affected by the flag
+    backend_a.get(backend_a.device, "rdu-spine1").model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_DST
+    diff = backend_a.diff_from(backend_a_minus_some_models)
+    assert diff.summary() == {"create": 0, "update": 0, "delete": 9, "no-change": 11}
+
+    # Check that only 1 object is affected by the flag
+    backend_a.get(backend_a.device, "sfo-spine2").model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_DST
+    diff = backend_a.diff_from(backend_a_minus_some_models)
+    assert diff.summary() == {"create": 0, "update": 0, "delete": 5, "no-change": 11}
```
glennmatthews, line 75: Comments each say "only 1 object" but it looks like the diffs have 3 objects "disappear" in the first case and four more objects "disappear" in the second case?

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def test_diffsync_sync_with_crud_no_update_flag_on_models(backend_a, exception_backend_a):
+    model1 = exception_backend_a.get(exception_backend_a.device, "nyc-spine1")
+    model1.role = "theotherrole"
+
+    with pytest.raises(NotImplementedError):
+        exception_backend_a.sync_from(backend_a)
+
+    model1.role = "theotherrole"
```
glennmatthews, line 120: Redundant with line 115?

diffsync/enum.py
```diff
@@ -35,6 +35,29 @@ class DiffSyncModelFlags(enum.Flag):
...
+
+    SKIP_UNMATCHED_DST = 0b1000
+    """Ignore the model if it only exists in the target/"to" DiffSync when determining diffs and syncing.
+
+    If this flag is set, the model will not be deleted from the target/"to" DiffSync.
+    """
+
+    SKIP_UNMATCHED_BOTH = SKIP_UNMATCHED_SRC | SKIP_UNMATCHED_DST
+
+    CRUD_NO_UPDATE = 0b10000
```
glennmatthews, line 52: Shouldn't we have a CRUD_NO_CREATE flag (to be set on the "source" side) as well? 

Relatedly, should we allow for CRUD_NO_UPDATE to be set on either the "source" or "target" model instances, and prevent changes in either case?

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,196 @@
...
+
+    model1 = exception_backend_a.get(exception_backend_a.device, "nyc-spine1")
+    model1.model_flags |= DiffSyncModelFlags.CRUD_NO_DELETE
+
+    diff = exception_backend_a.diff_from(backend_a)
+    assert diff.has_diffs()
+
+    exception_backend_a.sync_from(backend_a)
+    assert "nyc-spine1" in exception_backend_a.get(exception_backend_a.site, "nyc").devices
+    assert "nyc-spine1__eth0" not in exception_backend_a.get(exception_backend_a.device, "nyc-spine1").interfaces
```
glennmatthews, line 187: It's a little surprising to me that `CRUD_NO_DELETE` by itself doesn't prevent deletion of child objects. I can sort of understand requiring to also set `SKIP_CHILDREN_ON_DELETE` but it would be good to document this really clearly.

tests/unit/test_diffsync_model_flags.py
dgarros: This is because the interfaces of the device are also affected, I've updated the comment to be more explicit

tests/unit/test_diffsync_model_flags.py
dgarros: This behavior is inline with the behavior of the global flags `SKIP_UNMATCHED_*`. 
I've updated the unit tests for the global flags to confirm that.
I kind of agree with you but for now I think it's better to be consistent rather than introducing what could be considered a breaking change.

diffsync/enum.py
dgarros: I thought about it too but I'm not sure if this is possible to have flags defined on the source side currently because we are only passing the dest model to the `sync_model` method.

I think it would require us to retrieve the `src_model` from the `src_diffsync` and pass it to `sync_model()`. I can give it a try but if it turns out to be too complicated I might change my mind

diffsync/__init__.py
```diff
@@ -198,12 +197,32 @@ def create(cls, diffsync: "DiffSync", ids: Mapping, attrs: Mapping) -> Optional[
...
+            diffsync: The master data store for other DiffSyncModel instances that we might need to reference
+            ids: Dictionary of unique-identifiers needed to create the new object
+            attrs: Dictionary of additional attributes to set on the new object
+
+        Returns:
+            DiffSyncModel: instance of this class, if all data was successfully created.
+            None: if data creation failed in such a way that child objects of this model should not be created.
+
+        Raises:
+            ObjectNotCreated: if an error occurred.
```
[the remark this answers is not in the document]
dgarros: Good catch, I've removed the mention of `Returns: None...` and `Raises: ObjectNot*` and I mentioned the base methods in the doc of the other methods

tests/unit/test_diffsync_model_flags.py
dgarros: Doc update 
```
| CRUD_NO_DELETE | Do not delete and do not call delete() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff. This flag do not impact the children of this object, it's recommended to either set the flag on all objects individually or set the flag SKIP_CHILDREN_ON_DELETE | 0b1000000 |
```

diffsync/enum.py
dgarros: I've made the changes mentioned previously, I think it's cleaner overall

dgarros:
@glennmatthews I've made all the suggested changes and I've updated the doc, should be good for a final review, thanks

tests/unit/test_diffsync_model_flags.py
glennmatthews: Would it be possible and appropriate to add a "skipped" key to the diff summary? I'm okay with opening that as a separate enhancement request if it's non-trivial to implement.

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,236 @@
...
+    assert diff.summary() == {"create": 2, "update": 0, "delete": 0, "no-change": 23}
+
+    # Check that only 1 object is affected by the flag
+    backend_a_with_extra_models.get(
+        backend_a_with_extra_models.site, "lax"
+    ).model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_SRC
+    diff = backend_a.diff_from(backend_a_with_extra_models)
+    assert diff.summary() == {"create": 1, "update": 0, "delete": 0, "no-change": 23}
+
+    # Check that only the parent object is affected by the flag not the children
```
glennmatthews, line 36: Don't we need some diffs on the child objects for this to actually test what the comment is saying? Also, this statement seems at odds with the behavior of SKIP_UNMATCHED_DST (which does result in child objects being skipped as well).

tests/unit/test_diffsync_model_flags.py
```diff
@@ -0,0 +1,236 @@
...
+def test_diffsync_sync_with_skip_unmatched_dst_flag_on_models(backend_a, backend_a_minus_some_models):
+    backend_a.get(backend_a.site, "rdu").model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_DST
+    backend_a.get(backend_a.device, "sfo-spine2").model_flags |= DiffSyncModelFlags.SKIP_UNMATCHED_DST
+    backend_a.sync_from(backend_a_minus_some_models)
+
+    # Objects should not have been deleted
+    # rdu-spine1 hasn't been deleted because its parent hasn't been deleted
+    assert backend_a.get(backend_a.site, "rdu") is not None
+    assert backend_a.get(backend_a.device, "rdu-spine1") is not None
+    assert backend_a.get(backend_a.device, "sfo-spine2") is not None
```
glennmatthews, line 90: Also verify that sfo-spine2's child interfaces aren't deleted perhaps?

docs/source/core_engine/01-flags.md
```diff
@@ -57,6 +57,14 @@ class MyAdapter(DiffSync):
...
+| SKIP_UNMATCHED_BOTH | Convenience value combining both SKIP_UNMATCHED_SRC and SKIP_UNMATCHED_DST into a single flag | 0b1100 |
+| CRUD_NO_CREATE | Do not create and do not call create() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff. Because the destination model doesn't exist before calling create(), the flag need to be positioned on the source object. | 0b10000 |
+| CRUD_NO_UPDATE | Do not update and do not call update() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff. This flag can be positioned on both source and destination objects | 0b100000 |
+| CRUD_NO_DELETE | Do not delete and do not call delete() on the DiffSyncModel during sync(), the model and the changes will still be visible in the diff. This flag do not impact the children of this object, it's recommended to either set the flag on all objects individually or set the flag SKIP_CHILDREN_ON_DELETE | 0b1000000 |
```
glennmatthews, line 65: Looking at these flags side by side, it seems like there's some overlap. Is the only difference between `SKIP_UNMATCHED_DST` and `CRUD_NO_DELETE` that the former applies to both diff and sync while the latter applies only to sync? (If so, that would probably be helpful to state explicitly.) Same question for `SKIP_UNMATCHED_SRC` versus `CRUD_NO_CREATE`?

tests/unit/test_diffsync_model_flags.py
dgarros: Good idea, I created it https://github.com/networktocode/diffsync/issues/89

docs/source/core_engine/01-flags.md
dgarros: Yes that's correct, good point
it might be best to rename these flags
- `CRUD_NO_DELETE` > `SKIP_SYNC_UNMATCHED_DST`
- `CRUD_NO_CREATE` > `SKIP_SYNC_UNMATCHED_SRC`
- `CRUD_NO_UPDATE` > `SKIP_SYNC_UPDATE`
- `CRUD_NO_UPDATE_DELETE` change to `SKIP_SYNC_ALL` and add `SKIP_SYNC_UNMATCHED_SRC` to
what do you think ? happy to use different names if you have something better

diffsync/__init__.py
```diff
@@ -175,11 +175,29 @@ def set_status(self, status: DiffSyncStatus, message: Text = ""):
         self._status = status
         self._status_message = message
 
+    @classmethod
+    def create_base(cls, diffsync: "DiffSync", ids: Mapping, attrs: Mapping) -> Optional["DiffSyncModel"]:
```
glennmatthews, line 179: With the latest changes I'm back to not seeing why we need to have these `_base` methods added at all. If the subclass overrides `create()` then by default this method won't be called unless they call `super().create()`, which is now exactly identical to calling `super().create_base()` - so why have two different methods that do the same thing?

diffsync/helpers.py
```diff
@@ -339,42 +348,51 @@ def sync_diff_element(self, element: DiffElement, parent_model: "DiffSyncModel"
...
         for child in element.get_children():
-            changed |= self.sync_diff_element(child, parent_model=model)
+            changed |= self.sync_diff_element(child, parent_model=dst_model)
 
         return changed
 
-    def sync_model(
-        self, model: Optional["DiffSyncModel"], ids: Mapping, attrs: Mapping
+    def sync_model(  # pylint: disable=too-many-branches
+        self, src_model: Optional["DiffSyncModel"], dst_model: Optional["DiffSyncModel"], ids: Mapping, attrs: Mapping
```
glennmatthews, line 395: Looks like `src_model` is an unused variable now, so can we revert the changes to this function?

docs/source/core_engine/01-flags.md
```diff
@@ -57,6 +57,10 @@ class MyAdapter(DiffSync):
 |---|---|---|
 | IGNORE | Do not render diffs containing this model; do not make any changes to this model when synchronizing.  Can be used to indicate a model instance that exists but should not be changed by DiffSync. | 0b1 |
 | SKIP_CHILDREN_ON_DELETE | When deleting this model, do not recursively delete its children. Can be used for the case where deletion of a model results in the automatic deletion of all its children. | 0b10 |
+| SKIP_UNMATCHED_SRC | Ignore the model if it only exists in the source/"from" DiffSync when determining diffs and syncing. If this flag is set, no new model will be created in the target/"to" DiffSync. | 0b100 |
+| SKIP_UNMATCHED_DST | Ignore the model if it only exists in the target/"to" DiffSync when determining diffs and syncing. If this flag is set, the model will not be deleted from the target/"to" DiffSync. | 0b1000 |
+| SKIP_UNMATCHED_BOTH | Convenience value combining both SKIP_UNMATCHED_SRC and SKIP_UNMATCHED_DST into a single flag | 0b1100 |
+the diff. This flag do not impact the children of this object, it's recommended to either set the flag on all objects individually or set the flag SKIP_CHILDREN_ON_DELETE | 0b1000000 |
```
glennmatthews, line 63: Looks like some leftover text here?

diffsync/__init__.py
dgarros: I can remove them if you really want but I know if would have solved a situation for one of my project we had 3 layers of classes, like this 
1. DiffSyncModel
2. NetworkImporterModel
3. CustomerDefinedNetworkImporterModel

In this scenario, the goal was to redefine `NetworkImporterModel.update()` with `CustomerDefinedNetworkImporterModel.update()` to change it's behavior so calling super() wasn't possible but we still have to call `DiffSyncModel.update()`
We solved it by calling `DiffSyncModel.update()` directly but I think this isn't ideal either. 
I think., having the base method provides a clean way to identify the code that must be executed.

Again, happy to remove if you prefer ... I'm not gonna fall on my sword for this one

diffsync/helpers.py
dgarros: My preference would be to keep it in case we want to make a decision on a source flag in the future.
I think it was a miss in the design of this function

tests/unit/test_diffsync_model_flags.py
dgarros: Done

tests/unit/test_diffsync_model_flags.py
dgarros: Good point, the comment wasn't clear nor useful, I removed it

docs/source/core_engine/01-flags.md
dgarros: After more discussion we decided to remove all CRUD_* flags for now.
the same behavior can be achieve by setting different flags during a diff or a sync

diffsync/helpers.py
glennmatthews: Fair point.

diffsync/__init__.py
glennmatthews: I actually ran into something similar with the ssot-servicenow plugin and the `create()` method, where I didn't want to call the immediate superclass's `create()` but still needed to call the base `DiffSyncModel.create()`. This would have been potentially useful there as well; I'm just not entirely persuaded that needing to do this is a sign of a good code smell. :-)

glennmatthews (approved):

diffsync/__init__.py
dgarros: Agreed, I think it would be redesign the API for 2.0 to manage this situation better

glennmatthews (approved):

PR #90, opened 2022-01-07
Convert actions to Enum

### LINKED ISSUE #86: Create an Enum to track the valid type of Actions
### Environment
* DiffSync version: 1.3.0

### Proposed Functionality
Today, the list of valid actions : `create`, `update`, `delete` is not clearly defined and the value is hardcoded in multiple places in the code
It would be good to create a proper enum for these values and use it everywhere instead of having hardcoded values

### Use Case
House keeping

dgarros:
fixes #86 

Technically it's not a real enum because using when using an enum the value doesn't translate directly to a string but I think the result is the same.

### DIFF
--- a/diffsync/enum.py
+++ b/diffsync/enum.py
@@ -75,7 +75,7 @@ class DiffSyncStatus(enum.Enum):
     ERROR = "error"
 
 
-class DiffSyncActions:  # pylint: disable=too-few-public-methods
+class DiffSyncActions(enum.Enum):
     """List of valid Action for DiffSyncModel."""
 
     CREATE = "create"
--- a/diffsync/enum.py
+++ b/diffsync/enum.py
@@ -73,3 +73,12 @@ class DiffSyncStatus(enum.Enum):
     SUCCESS = "success"
     FAILURE = "failure"
     ERROR = "error"
+
+
+class DiffSyncActions:  # pylint: disable=too-few-public-methods
+    """List of valid Action for DiffSyncModel."""
+
+    CREATE = "create"
+    UPDATE = "update"
+    DELETE = "delete"
+    NO_CHANGE = None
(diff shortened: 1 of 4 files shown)


diffsync/enum.py
```diff
@@ -73,3 +73,12 @@ class DiffSyncStatus(enum.Enum):
     SUCCESS = "success"
     FAILURE = "failure"
     ERROR = "error"
+
+
+class DiffSyncActions:  # pylint: disable=too-few-public-methods
```
glennmatthews, line 78: I don't like having a "not a real Enum" class in `enum.py`. Can you provide more context as to why we're not using an actual `enum.Enum` class here, unlike all the other classes in this file?

dgarros: I guess it was my inexperience with the enum, somehow I expected them to behave like the `ChoiceSet` in Nautobot and to translate directly to the real value
```
>>> DiffSyncActions.CREATE
"create"
``` 

I guess the right solution is just to use `DiffSyncActions.CREATE.value` instead but honestly it's not 100% clear to me what is the value to use the real enum class instead of a simple dataclass.

diffsync/helpers.py
```diff
@@ -390,27 +390,27 @@ def sync_model(
...
                     raise ObjectNotUpdated(f"Failed to update {self.model_class.get_type()} {ids} - not found!")
                 model = model.update(attrs=attrs)
-            elif self.action == "delete":
+            elif self.action == DiffSyncActions.DELETE:
                 if model is None:
                     raise ObjectNotDeleted(f"Failed to delete {self.model_class.get_type()} {ids} - not found!")
                 model = model.delete()
             else:
-                raise ObjectCrudException(f'Unknown action "{self.action}"!')
+                raise ObjectCrudException(f'Unknown action "{self.action.value}"!')
```
glennmatthews, line 407: I'd probably keep this as `self.action` rather than `self.action.value` just in case we get something in that isn't even a `DiffSyncActions`?

glennmatthews (approved):

dgarros:
@glennmatthews I've updated this PR to use a proper Enum, ready to review

diffsync/helpers.py
dgarros: Good point

glennmatthews (approved):

PR #95, opened 2022-01-26
Remote extra docs from pyproject and fix dependencies and prep for 1.4.1

### LINKED ISSUE #94: Pydantic and structlog are not installed by default
### Environment
* DiffSync version:  1.4.0
* Python version  All
 
### Observed Behavior
When installing DiffSync from pip, some mandatory dependencies like pydantic or structlog are not being installed automatically

### Expected Behavior
All mandatory dependencies should be installed by default

dgarros:
Fixes #94 

After looking more closely at the `pyproject.toml` it felt like the extra `[docs]` wasn't required so I removed it and I moved back all docs dependencies to `dev-dependencies`. Some of them were already in there anyway.

Also in a matter of time, I increased the package version to 1.4.1

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,17 +2,17 @@
 name = "alabaster"
 version = "0.7.12"
 description = "A configurable sidebar-enabled Sphinx theme"
-category = "main"
+category = "dev"
 optional = false
 python-versions = "*"
 
 [[package]]
 name = "astroid"
-version = "2.9.0"
+version = "2.9.3"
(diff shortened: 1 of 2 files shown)


glennmatthews:
Code changes look fine but I don't see any CI result for ReadTheDocs. Do we not have it enabled yet on this repo?

dgarros:
@glennmatthews I'll have to check, not sure if it's automatically building the doc when we are pushing some changes to main
I'll open another issue to track that

glennmatthews (approved):

PR #102, opened 2022-03-02
Convert DiffSyncActions enum into simple string

### LINKED ISSUE #101: Breaking API change in DiffSync 1.4.0
### Environment
* DiffSync version:  1.4.0
* Python version: any

### Observed Behavior

#90 introduced a breaking API change in that `DiffElement.action` changed from a string value to an enum value. This impacts projects such as `network-importer` (https://github.com/networktocode/network-importer/issues/256) and any other project that is relying on the value of `DiffElement.action`, such as to implement custom Diff ordering based on the action. 

In the short term it may be simplest just to revert the entirety of #90 and cut a new DiffSync release.

### Expected Behavior

API to remain stable in minor and patch releases.

### Steps to Reproduce
1. See for example https://github.com/networktocode/network-importer/issues/256

dgarros:
Fixes #101

### DIFF
--- a/diffsync/diff.py
+++ b/diffsync/diff.py
@@ -106,9 +106,9 @@ class Diff:
     def summary(self) -> Mapping[Text, int]:
         """Build a dict summary of this Diff and its child DiffElements."""
         summary = {
-            DiffSyncActions.CREATE.value: 0,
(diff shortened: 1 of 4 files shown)


glennmatthews (approved):
Thank you!

PR #103, opened 2022-03-02
Update development dependencies

dgarros:
This update update all the dev dependencies to the latest version

I couldn't install some dev dependencies  on Python 3.9 because the lock file what pointing at a specific version of `typed-ast` that isn't support on Mac M1

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -69,30 +69,26 @@ stevedore = ">=1.20.0"
 
 [[package]]
 name = "black"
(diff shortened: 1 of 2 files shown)


glennmatthews (approved):
Approved with reservations - I feel like `dependency = "*"` is asking for trouble in the future, even for dev-only dependencies.

dgarros:
thanks

PR #106, opened 2022-04-13
Implement a Redis backend storage

chadell:
This PRs addresses issue #57 partially, abstracting the store for DiffSyncModel objects, and extending to using Redis  backend besides the local memory.

This feature it's specially important when working on a distributed data pipeline when different workers would load the data from adapters and finally the data must be shared to create the diff. It leverages on #58 proposal.

The implementation  approach has moved the previous `_data` into the `LocalStore` implementation, creating abstract methods to interact with the storage, so we were able to create another implementation using Redis.

# TODO

- [x] Extend Redis test coverage and refine cornercase
- [x] Fix basic testing
- [x] Update Docs (to be done when final approach is agreed)

### DIFF
--- a/diffsync/store/__init__.py
+++ b/diffsync/store/__init__.py
@@ -103,7 +103,7 @@ class BaseStore:
         """
         raise NotImplementedError
 
-    def count(self, *, modelname=None) -> int:
+    def count(self, *, model: Union[Text, "DiffSyncModel", Type["DiffSyncModel"], None] = None) -> int:
         """Returns the number of elements of a specific model, or all elements in the store if not specified."""
         raise NotImplementedError
 
--- a/diffsync/store/__init__.py
+++ b/diffsync/store/__init__.py
@@ -1,0 +1,247 @@
+from collections import defaultdict
+from typing import Callable, ClassVar, Dict, List, Mapping, MutableMapping, Optional, Text, Tuple, Type, Union
+
+from diffsync.exceptions import ObjectNotFound, ObjectAlreadyExists
+
+
+class BaseStore:
+
+    def __init__(self, diffsync=None, name=None, *args, **kwargs) -> None:
+        self.diffsync = diffsync
+        self.name = name if name else self.__class__.__name__
+
+    def __str__(self):
+        return self.name
+
+    def get(self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]], identifier: Union[Text, Mapping]):
+        raise NotImplementedError
+
+    def get_all(self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]]) -> List["DiffSyncModel"]:
+        raise NotImplementedError
+
+    def get_by_uids(
+        self, uids: List[Text], obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]]
+    ) -> List["DiffSyncModel"]:
+        raise NotImplementedError
+
+    def remove(self, obj: "DiffSyncModel", remove_children: bool = False):
+        raise NotImplementedError
+
+    def add(self, obj: "DiffSyncModel"):
+        raise NotImplementedError
+
+    def update(self, obj: "DiffSyncModel"):
+        raise NotImplementedError
+
+    def count(self, modelname):
+        raise NotImplementedError
+
+    def get_or_instantiate(
+        self, model: Type["DiffSyncModel"], ids: Dict, attrs: Dict = None
+    ) -> Tuple["DiffSyncModel", bool]:
+        """Attempt to get the object with provided identifiers or instantiate it with provided identifiers and attrs.
+
+        Args:
+            model (DiffSyncModel): The DiffSyncModel to get or create.
+            ids (Mapping): Identifiers for the DiffSyncModel to get or create with.
+            attrs (Mapping, optional): Attributes when creating an object if it doesn't exist. Defaults to None.
+
+        Returns:
+            Tuple[DiffSyncModel, bool]: Provides the existing or new object and whether it was created or not.
+        """
+        created = False
+        try:
+            obj = self.get(model, ids)
+        except ObjectNotFound:
+            if not attrs:
+                attrs = {}
+            obj = model(**ids, **attrs)
+            # Add the object to diffsync adapter
+            self.add(obj)
+            created = True
+
+        return obj, created
+
+    def update_or_instantiate(
+        self, model: Type["DiffSyncModel"], ids: Dict, attrs: Dict
+    ) -> Tuple["DiffSyncModel", bool]:
+        """Attempt to update an existing object with provided ids/attrs or instantiate it with provided identifiers and attrs.
+
+        Args:
+            model (DiffSyncModel): The DiffSyncModel to get or create.
+            ids (Dict): Identifiers for the DiffSyncModel to get or create with.
+            attrs (Dict): Attributes when creating/updating an object if it doesn't exist. Pass in empty dict, if no specific attrs.
+
+        Returns:
+            Tuple[DiffSyncModel, bool]: Provides the existing or new object and whether it was created or not.
+        """
+        created = False
+        try:
+            obj = self.get(model, ids)
+        except ObjectNotFound:
+            obj = model(**ids, **attrs)
+            # Add the object to diffsync adapter
+            self.add(obj)
+            created = True
+
+        # Update existing obj with attrs
+        for attr, value in attrs.items():
+            if getattr(obj, attr) != value:
+                setattr(obj, attr, value)
+
+        return obj, created
+
+
+class LocalStore(BaseStore):
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self._data = defaultdict(dict)
+
+    def get(self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]], identifier: Union[Text, Mapping]):
+        """Get one object from the data store based on its unique id.
+
+        Args:
+            obj: DiffSyncModel class or instance, or modelname string, that defines the type of the object to retrieve
+            identifier: Unique ID of the object to retrieve, or dict of unique identifier keys/values
+
+        Raises:
+            ValueError: if obj is a str and identifier is a dict (can't convert dict into a uid str without a model class)
+            ObjectNotFound: if the requested object is not present
+        """
+        if isinstance(obj, str):
+            modelname = obj
+            if not hasattr(self, obj):
+                object_class = None
+            else:
+                object_class = getattr(self, obj)
+        else:
+            object_class = obj
+            modelname = obj.get_type()
+
+        if isinstance(identifier, str):
+            uid = identifier
+        elif object_class:
+            uid = object_class.create_unique_id(**identifier)
+        else:
+            raise ValueError(
+                f"Invalid args: ({obj}, {identifier}): "
+                f"either {obj} should be a class/instance or {identifier} should be a str"
+            )
+
+        if uid not in self._data[modelname]:
+            raise ObjectNotFound(f"{modelname} {uid} not present in {str(self)}")
+        return self._data[modelname][uid]
+
+    def get_all(self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]]) -> List["DiffSyncModel"]:
+        """Get all objects of a given type.
+
+        Args:
+            obj: DiffSyncModel class or instance, or modelname string, that defines the type of the objects to retrieve
+
+        Returns:
+            List[DiffSyncModel]: List of Object
+        """
+        if isinstance(obj, str):
+            modelname = obj
+        else:
+            modelname = obj.get_type()
+
+        return list(self._data[modelname].values())
+
+    def get_by_uids(
+        self, uids: List[Text], obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]]
+    ) -> List["DiffSyncModel"]:
+        """Get multiple objects from the store by their unique IDs/Keys and type.
+
+        Args:
+            uids: List of unique id / key identifying object in the database.
+            obj: DiffSyncModel class or instance, or modelname string, that defines the type of the objects to retrieve
+
+        Raises:
+            ObjectNotFound: if any of the requested UIDs are not found in the store
+        """
+        if isinstance(obj, str):
+            modelname = obj
+        else:
+            modelname = obj.get_type()
+
+        results = []
+        for uid in uids:
+            if uid not in self._data[modelname]:
+                raise ObjectNotFound(f"{modelname} {uid} not present in {str(self)}")
+            results.append(self._data[modelname][uid])
+        return results
+
+    def add(self, obj: "DiffSyncModel"):
+        """Add a DiffSyncModel object to the store.
+
+        Args:
+            obj (DiffSyncModel): Object to store
+
+        Raises:
+            ObjectAlreadyExists: if a different object with the same uid is already present.
+        """
+        modelname = obj.get_type()
+        uid = obj.get_unique_id()
+
+        existing_obj = self._data[modelname].get(uid)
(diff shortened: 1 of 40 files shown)


chadell:
@glennmatthews I found an issue with this PR, specifically on the handling of `children`. Today we run the `add_child` and `remove_child` directly in the `DiffSyncModel`, and this works well when we run it in memory, but when using another store, such as Redis, this changes in the memory object are not update to the backend store.

For instance, from [example 01](https://github.com/networktocode/diffsync/blob/main/examples/01-multiple-data-sources/backend_a.py):

```python
...
                self.add(device)
                site.add_child(device)
...
```

In this example, `site` is no longer updated in the store. We would need to to a `self.update(site)` after this change, but I'm concerned on the backwards compatibility implications this may have.

Obviously, this would be only mandatory when using a non local memory store, but to reuse the code in the other cases we would need to change this part of the code.

Looking forward to your thoughts

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(self, *args, store_id=None, host="localhost", port=6379, url=None, db=0, **kwargs):
+        """Init method for RedisStore."""
+        super().__init__(*args, **kwargs)
+
+        try:
+            if url:
+                self._store = Redis.from_url(url, db=db)
+            else:
+                self._store = Redis(host=host, port=port, db=db)
```
glennmatthews, line 29: I'm not a big fan of having both `url` and `host/port` as mutually exclusive arguments. Can we converge to just one or the other? If not, can we have some error checking to enforce that they actually are mutually exclusive?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+            if not self._store.ping():
+                raise RedisConnectionError
+        except RedisConnectionError:
+            raise ObjectStoreException("Redis store is unavailable.") from RedisConnectionError
+
+        self._store_id = store_id if store_id else str(uuid.uuid4())[:8]
```
glennmatthews, line 36: Why only the first 8 characters of the UUID?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+        self._store_label = f"{REDIS_DIFFSYNC_ROOT_LABEL}:{self._store_id}"
+
+    def __str__(self):
+        """Render store name."""
+        return f"{self.name}({self._store_id})"
```
glennmatthews, line 42: Perhaps?
```suggestion
        return f"{self.name} ({self._store_id})"
```

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def get_all_model_names(self):
```
glennmatthews, line 44:
```suggestion
    def get_all_model_names(self) -> List[str]:
```

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+        """Get all the model names stored.
+
+        Return:
+            List[str]: List of all the model names.
+        """
+        # TODO: optimize it
+        all_model_names = []
+        for item in self._store.scan_iter(f"{self._store_label}:*"):
+            model_name = item.split(b":")[2].decode()
```
glennmatthews, line 53: A comment here explaining this logic would be appreciated.

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __str__(self):
+        """Render store name."""
+        return f"{self.name}({self._store_id})"
+
+    def get_all_model_names(self):
+        """Get all the model names stored.
+
+        Return:
+            List[str]: List of all the model names.
```
glennmatthews, line 48: Looking at the implementation here, I wonder if it would be more appropriate for `get_all_model_names` to return a Set rather than a List, since its members should be unique and order is not significant.

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+            model_name = item.split(b":")[2].decode()
+            if model_name not in all_model_names:
+                all_model_names.append(model_name)
+
+        return all_model_names
+
+    def _get_key_for_object(self, modelname, uid):
+        return f"{self._store_label}:{modelname}:{uid}"
+
+    def get(self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]], identifier: Union[Text, Mapping]):
```
glennmatthews, line 62:
```suggestion
    def get(
        self, obj: Union[Text, "DiffSyncModel", Type["DiffSyncModel"]], identifier: Union[Text, Mapping]
    ) -> "DiffSyncModel":
```

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+        if isinstance(identifier, str):
+            uid = identifier
+        elif object_class:
+            uid = object_class.create_unique_id(**identifier)
+        else:
+            raise ValueError(
+                f"Invalid args: ({obj}, {identifier}): "
+                f"either {obj} should be a class/instance or {identifier} should be a str"
+            )
```
glennmatthews, line 91: This logic is the same across Store classes - consider creating a helper method on `BaseStore` that they could both call?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
+"""RedisStore module."""
+import copy
+import uuid
+from pickle import loads, dumps  # nosec
```
glennmatthews, line 4: Since DiffSyncModels are Pydantic classes, it shouldn't be necessary to use `pickle` if security is a concern - you should be able to instead use `model.json()` to dump to JSON, and `<model_class>.parse_raw()` to instantiate it from JSON. Not sure how comparable the performance will be versus pickle, but worth considering?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+            raise ValueError(
+                f"Invalid args: ({obj}, {identifier}): "
+                f"either {obj} should be a class/instance or {identifier} should be a str"
+            )
+
+        try:
+            obj_result = loads(self._store.get(self._get_key_for_object(modelname, uid)))  # nosec
+            obj_result.diffsync = self.diffsync
+        except TypeError:
+            raise ObjectNotFound(f"{modelname} {uid} not present in Cache")  # pylint: disable=raise-missing-from
```
glennmatthews, line 97: Why `disable=raise-missing-from` instead of just using `from exc` here and below?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+            modelname = obj.get_type()
+
+        results = []
+        for key in self._store.scan_iter(f"{self._store_label}:{modelname}:*"):
+            try:
+                obj_result = loads(self._store.get(key))  # nosec
+                obj_result.diffsync = self.diffsync
+                results.append(obj_result)
+            except TypeError:
+                raise ObjectNotFound(f"{key} not present in Cache")  # pylint: disable=raise-missing-from
```
glennmatthews, line 122: Consider creating a helper method for this common pattern?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+        if remove_children:
+            for child_type, child_fieldname in obj.get_children_mapping().items():
+                for child_id in getattr(obj, child_fieldname):
+                    try:
+                        child_obj = self.get(child_type, child_id)
+                        self.remove(child_obj, remove_children=remove_children)
+                    except ObjectNotFound:
+                        pass
+                        # Since this is "cleanup" code, log an error and continue, instead of letting the exception raise
+                        # self._log.error(f"Unable to remove child {child_id} of {modelname} {uid} - not found!")
```
glennmatthews, line 234: Why commented out?

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def count(self, modelname=None):
```
glennmatthews, line 236:
```suggestion
    def count(self, modelname=None) -> int:
```

diffsync/store/redis.py
```diff
@@ -0,0 +1,242 @@
...
+        """Returns the number of elements of an specific model name."""
+        search_pattern = f"{self._store_label}:*"
+        if modelname:
+            search_pattern = f"{self._store_label}:{modelname.lower()}:*"
+
+        return sum([1 for _ in self._store.scan_iter(search_pattern)])
```
glennmatthews, line 242: Can we not do something like:

```suggestion
        return len(self._store.scan_iter(search_pattern))
```

pyproject.toml
```diff
@@ -23,6 +23,8 @@ packaging = "^21.3"
 colorama = {version = "^0.4.3", optional = true}
 # For Pydantic
 dataclasses = {version = "^0.7", python = "~3.6"}
+redis = "*"
+types-redis = "*"
```
glennmatthews, line 27: Should `types-redis` be a dev-only dependency, like `types-toml`?

pyproject.toml
```diff
@@ -23,6 +23,8 @@ packaging = "^21.3"
 colorama = {version = "^0.4.3", optional = true}
 # For Pydantic
 dataclasses = {version = "^0.7", python = "~3.6"}
+redis = "*"
```
glennmatthews, line 26: 1. Is there a known minimum version of `redis` that's required?
2. Can we make this into an optional/extras dependency, i.e. so that `pip install diffsync` doesn't pull in redis (and only supports local storage as a result), but `pip install diffsync[redis]` does?

tests/unit/test_diff_element.py
```diff
@@ -33,7 +33,7 @@ def test_diff_element_empty():
     assert not element.has_diffs()
     assert not element.has_diffs(include_children=True)
     assert not element.has_diffs(include_children=False)
-    assert element.get_attrs_keys() == []
+    assert not element.get_attrs_keys()
```
glennmatthews, line 36: Revert please.

tests/unit/test_diffsync.py
```diff
@@ -737,7 +738,9 @@ def test_diffsync_remove_missing_child(log, backend_a):
     backend_a.remove(rdu_spine1_eth0)
     # Should log an error but continue removing other child objects
     backend_a.remove(rdu_spine1, remove_children=True)
-    assert log.has("Unable to remove child rdu-spine1__eth0 of device rdu-spine1 - not found!", diffsync=backend_a)
+    assert log.has(
+        "Unable to remove child rdu-spine1__eth0 of device rdu-spine1 - not found!", diffsync=backend_a.store
```
glennmatthews, line 742: This seems wrong - `diffsync` parameter should be the backend, not the store

diffsync/store/redis.py
chadell: The url was added to be able to use the test emulation library for Redis. I will enforce the checking.

diffsync/store/redis.py
chadell: added

pyproject.toml
chadell: good idea

diffsync/store/redis.py
```diff
@@ -18,10 +18,13 @@
 class RedisStore(BaseStore):
     """RedisStore class."""
 
-    def __init__(self, *args, store_id=None, host="localhost", port=6379, url=None, db=0, **kwargs):
+    def __init__(self, *args, store_id=None, host=None, port=6379, url=None, db=0, **kwargs):
         """Init method for RedisStore."""
         super().__init__(*args, **kwargs)
 
+        if url and host and port:
```
glennmatthews, line 25: Perhaps?

```suggestion
        if url and (host or port):
```

diffsync/store/redis.py
```diff
@@ -226,17 +208,21 @@ def remove(self, obj: "DiffSyncModel", remove_children: bool = False):
...
                         # Since this is "cleanup" code, log an error and continue, instead of letting the exception raise
                         # self._log.error(f"Unable to remove child {child_id} of {modelname} {uid} - not found!")
 
-    def count(self, modelname=None):
-        """Returns the number of elements of an specific model name."""
+    def count(self, *, model: Union[Text, "DiffSyncModel", Type["DiffSyncModel"], None] = None) -> int:
+        """Returns the number of elements of a specific model, or all elements in the store if unspecified."""
         search_pattern = f"{self._store_label}:*"
-        if modelname:
+        if model:
```
glennmatthews, line 196: I think I remember some bug in the past where either a DiffSyncModel or a DiffSync was evaluating as falsey, resulting in unexpected behavior. Perhaps:

```suggestion
        if model is not None:
```

diffsync/store/redis.py
chadell: if we do like this, then I have to remove the default port from the kwargs, that I thought it would be convenient. But, it could be also confusing... not sure

diffsync/store/redis.py
chadell: I was not aware of the `parse_raw()` helper method in pydantic models. Good one.
However, I have a question. If I use the `model.json()` method I'm losing some fields vs `dict()`, that are not simple to convert to json:

```
(Pdb) obj.json()
'{"name": "site1"}'
(Pdb) obj.dict()
{'model_flags': <DiffSyncModelFlags.NONE: 0>, 'name': 'site1', 'devices': []}
```

This was the reason to use pickle, and get the raw objects directly. Do you see a proper way to overcome that?

glennmatthews: That's surprising to me that `json()` is discarding properties from `dict()`, I wouldn't have expected that from reading the Pydantic docs. I'm fine with sticking with `pickle` if the baseline json/dict serialization is not adequate to the task.

diffsync/store/redis.py
glennmatthews: OK, perhaps just `if url and host` then?

diffsync/store/__init__.py
```diff
@@ -72,6 +72,10 @@ def get_by_uids(
         """
         raise NotImplementedError
 
+    def _remove_item(self, modelname: str, uid: str):
```
glennmatthews, line 75: I don't think I like having a private but abstract method. If this is something subclasses are responsible for implementing, it shouldn't be private, IMHO.

diffsync/store/__init__.py
```diff
@@ -82,7 +86,23 @@ def remove(self, *, obj: "DiffSyncModel", remove_children: bool = False):
...
+
+        if remove_children:
+            for child_type, child_fieldname in obj.get_children_mapping().items():
+                for child_id in getattr(obj, child_fieldname):
+                    try:
+                        child_obj = self.get(model=child_type, identifier=child_id)
+                        self.remove(obj=child_obj, remove_children=remove_children)
+                    except ObjectNotFound:
+                        # Since this is "cleanup" code, log an error and continue, instead of letting the exception raise
+                        self._log.error(f"Unable to remove child {child_id} of {modelname} {uid} - not found!")
```
glennmatthews, line 105: It occurs to me we should probably be taking advantage of structlog better on this error message. Perhaps something like:

```suggestion
                        self._log.error(f"Unable to remove child element as it was not found!", child_type=child_type, child_id=child_id, parent_type=modelname, parent_id=uid)
```

diffsync/store/redis.py
chadell: I guess that the values that can't be serialized, are discarded.

diffsync/store/redis.py
```diff
@@ -0,0 +1,203 @@
+"""RedisStore module."""
+import copy
+import uuid
+from pickle import loads, dumps  # nosec
+from typing import List, Mapping, Text, Type, Union, TYPE_CHECKING, Set
+
+from redis import Redis
+from redis.exceptions import ConnectionError as RedisConnectionError
```
glennmatthews, line 8: Don't we need some logic here to handle the case where `redis` isn't installed?

diffsync/store/__init__.py
chadell: yes, I remember a previous conversation about this public/private question :)

glennmatthews (approved):

PR #111, opened 2022-06-09
New Example06 - IP Prefixes

chadell:


examples/06-ip-prefixes/README.md
```diff
@@ -0,0 +1,122 @@
...
+## Test the example
+
+You could simply run the `main.py` file, but to run step by step.
+
+### Set up the environment
+
+Install the dependencies (recommended into a virtual environment)
+
+```
+pip3 install -r requirements
```
glennmatthews, line 16:
```suggestion
pip3 install -r requirements.txt
```

examples/06-ip-prefixes/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+### Initialize and load adapter for IPAM A
+
+```py
+>>> ipam_a = IpamA()
+>>> ipam_a.load()
+```
+
+You can check the content loaded from IPAM A. Notice that the data has been transformed into the DiffSync model, that is different from the original one.
```
glennmatthews, line 40:
```suggestion
You can check the content loaded from IPAM A. Notice that the data has been transformed into the DiffSync model, which is different from the original YAML data.
```

examples/06-ip-prefixes/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+### Initialize and load adapter for IPAM B
+
+```py
+>>> ipam_b = IpamB()
+>>> ipam_b.load()
+```
+
+You can check the content loaded from IPAM B. Notice that the data has been transformed into the DiffSync model, that is different from the original one.
```
glennmatthews, line 62:
```suggestion
You can check the content loaded from IPAM B. Notice that the data has been transformed into the DiffSync model, which again is different from the original YAML format.
```

examples/06-ip-prefixes/README.md
```diff
@@ -0,0 +1,122 @@
...
+{'prefix': {'10.10.10.10/24': {'prefix': '10.10.10.10/24', 'vlan_id': 123},
+            '2001:DB8::/32': {'prefix': '2001:DB8::/32',
+                              'tenant': 'XYZ Corporation',
+                              'vlan_id': 10,
+                              'vrf': 'data'}}}
+```
+
+### Check the difference
+
+We can use `diff_to` or `diff_sync` to select, from the perspective of the calling adapter, who is the authoritative in each case.
```
glennmatthews, line 75:
```suggestion
We can use `diff_to` or `diff_from` to select, from the perspective of the calling adapter, who is the authoritative in each case.
```

examples/06-ip-prefixes/README.md
```diff
@@ -0,0 +1,122 @@
...
+### Validate synchronization
+
+Now, if we reload the IPAM B, and try to check the difference, we should see no differences.
+
+```py
+>>> new_ipam_b = IpamB().load()
+>>> diff = ipam_a.diff_to(ipam_b)
+>>> diff.summary()
+{'create': 0, 'update': 0, 'delete': 0, 'no-change': 3}
+```
```
glennmatthews, line 122: Something seems off here - you're loading `new_ipam_b` but then checking the diff between `ipam_a` and `ipam_b`?

examples/06-ip-prefixes/adapter_ipam_a.py
```diff
@@ -0,0 +1,94 @@
...
+            {
+                "cidr": ids["prefix"],
+                "family": ipaddress.ip_address(ids["prefix"].split("/")[0]).version,
+                "vrf": attrs["vrf"],
+                "vlan": f'VLAN{attrs["vlan_id"]}',
+                "customer_id": attrs["tenant"] if attrs["tenant"] else None,
+            }
+        )
+
+        diffsync.write_yaml_data(data)
```
glennmatthews, line 29: Seems wrong to me to re-load and re-dump the YAML file after each individual model create/update/delete. Should this be done (once) in a `sync_complete` callback instead?

chadell: you are totally right

glennmatthews (approved):

PR #114, opened 2022-06-29
Make an empty DiffSync instance evaluate to True.

### LINKED ISSUE #84: Empty DiffSync instance evaluates as false
### Environment
* DiffSync version:  1.3.0
* Python version:  3.6

### Observed Behavior
When using diffsync within Nautobot-plugin-ssot, it would be nice to allow the Destination (Target) to have no data.

An empty DiffSync() evaluates to False. This causes a failure to proceed to the Diff.

the base DiffSync class has a __len__() method, an object’s default bool() casting must use that to determine whether it evaluates as truthy:
>>> diffsync = DiffSync()
>>> bool(diffsync)
False
>>> diffsync.add(DiffSyncModel())
>>> bool(diffsync)
True

### Expected Behavior
diffsync proceeds to diff and then syncs data to the Destination (Target).

Kircheneer:
Fixes #84


Kircheneer:
As discussed this really should warrant a major version increase and as such will have to be delayed until we have more breaking changes.

PR #115, opened 2022-06-29
Attempt to fix the read the docs pipeline.

### LINKED ISSUE #96: Fix build pipeline for Read The Doc
### Environment
* DiffSync version:  1.4.1
* Python version  3.7

### Observed Behavior
The build pipeline for Read The Doc appears to be broken at the moment
The PR #95 didn't help for sure but I think it was broken even before that.

Looking at other projects, it doesn't look like we have a clear pattern in place but I think we should replicate that we have in netutils with a dedicated `requirements.txt` file, just for RTD 
https://github.com/networktocode/netutils/blob/develop/docs/requirements.txt

### Expected Behavior
A new version of the documentation should be build and published to RTD when we have a new commit in `main`

Kircheneer:
Fixes #96

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,9 +7,6 @@ sphinx:
   fail_on_warning: false
 
 python:
-  version: 3.7
+  version: "3.7"
   install:
-    - method: "pip"
-      path: "."
-      extra_requirements:
-        - "docs"
+    - requirements: docs/requirements.txt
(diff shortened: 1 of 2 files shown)


Kircheneer:
Working, see here https://diffsync-test-leo.readthedocs.io/en/latest/

glennmatthews (approved):

PR #127, opened 2022-07-12
Initial draft for field filtering.

Kircheneer:


itdependsnetworks:
This is pretty cool and has come up multiple times so appreciate it. I do think we need to consider the name of the feature, perhaps something like "limit fields", as reading this, I thought it was about limiting the scope of data.

diffsync/__init__.py
```diff
@@ -71,6 +72,20 @@ class DiffSyncModel(BaseModel):
     Note: inclusion in `_attributes` is mutually exclusive from inclusion in `_identifiers`; a field cannot be in both!
     """
 
+    _enabled_attributes: ClassVar[Mapping[str, bool]] = defaultdict(lambda: True)
+    """Optional: dict of `{field_name: enabled}` entries detailling whether an attribute is enabled or not.
+    
+    When a field is not included in this dictionary, it means it is enabled (this also provides backwards
+    compatibility). When a field is
+    - included as `_enabled_attributes[field_name] == False`, this means that while the attribute has to defined like
+      any other, it will not take part in any diffsync logic
```
glennmatthews, line 81: Hmm. Probably I'm missing something, but can't the same effect be achieved with current functionality by simply overriding `_attributes` on the concrete subclass?

chadell: I see it like @glennmatthews 🤔 , maybe we need more context?

Kircheneer: It appears you are correct and I have gone full circle 🤦🏻

Kircheneer:
Closing in favor of #142 - it appears that no code change is needed in order to accomplish this.

PR #141, opened 2022-07-26
Update renovate.json

Kircheneer:
Currently renovate appears to be using poetry 1.1.13, which is a problem because of [this](https://github.com/python-poetry/poetry/issues/6045). According to [this](https://github.com/python-poetry/poetry/issues/6045), we can apply constraints to the Poetry version used by renovate with the change I did here.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,15 +2,13 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "baseBranches": ["develop"],
   "constraints": {
-    "python": "3.7"
+    "python": ">=3.7",
+    "poetry": ">=1.1.14"
   },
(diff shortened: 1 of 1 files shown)


glennmatthews:
Any way to prove that this does what it says? :-)

Kircheneer:
There is the way of "install the whole renovate stack locally and see what it does", I was looking to avoid that but should probably go do it..

Kircheneer:
Check [this](https://github.com/Kircheneer/diffsync/pull/4) for proof @glennmatthews 👍🏻

Kircheneer:
@glennmatthews @chadell can we move this forward? I'd like to get the whole renovate configuration thing wrapped up 🚀

glennmatthews (approved):

PR #144, opened 2022-07-27
Introduce cruft to keep the repository up to date with template changes

### LINKED ISSUE #139: Re-evaluate pytest-redislite as a dependency
# Problem

The testing dependency pytest-redislite depends on [redislite](https://github.com/yahoo/redislite) which brings with it a couple of problems:
- No version that installs on M1 Macs out of the box, there are some [workarounds](https://github.com/yahoo/redislite/issues/152) you can perform but as per my experience they don't play too well with poetry
- Latest release is about 1.5 years in the past
- General [signs of abandonment](https://github.com/yahoo/redislite/issues/164) in the project, although that issue was closed the issues inside of it weren't really addressed yet

# Potential solutions

- Require docker-compose to run tests with Redis since we always have docker running anyway for development and then do away with mocking Redis, use [pytest-redis](https://github.com/ClearcodeHQ/pytest-redis)?
- Switch to something else for mocking Redis such as [fakeredis-py](https://github.com/dsoftwareinc/fakeredis-py)?

Kircheneer:
- Introduce cruft
- Replace `pytest-redislite` with `pytest-redis` as a consequence unit tests needs to be run in Docker or redis needs to be installed locally (fixes #139 and enable Docker buildx to function)

Why is this not in the CI (already)? I wasn't sure about including it because that would mean that updates to the cookiecutter template would cause this CI to somewhat arbitrarily fail, we should follow up at a later point, maybe post diffs to slack if there are any?


chadell (commented):
awesome! 🎉

chadell (dismissed):
LGTM

.github/workflows/local.yml
```diff
@@ -0,0 +1,29 @@
...
+  mypy:
+    runs-on: "ubuntu-20.04"
+    env:
+      INVOKE_LOCAL: "True"
+    steps:
+      - name: "Check out repository code"
+        uses: "actions/checkout@v2"
+      - name: "Setup environment"
+        uses: "networktocode/gh-action-setup-poetry-environment@v1"
+      - name: "Linting: flake8"
```
glennmatthews, line 28:
```suggestion
      - name: "Linting: mypy"
```

.github/workflows/local.yml
```diff
@@ -0,0 +1,29 @@
+---
```
glennmatthews, line 1: Is this a separate file just so as to keep `ci.yml` identical with the cookiecutter? I would have hoped that it'd be possible/straightforward with `cruft` to have a locally modified version of the file (i.e. some consistent diff against the baseline). This feels like a bit of a step backwards to me.

.gitignore
```diff
@@ -183,11 +192,48 @@ $RECYCLE.BIN/
...
+.idea/**/dataSources.ids
+.idea/**/dataSources.local.xml
+.idea/**/sqlDataSources.xml
+.idea/**/dynamic.xml
+.idea/**/uiDesigner.xml
+.idea/**/dbnavigator.xml
+
+# Gradle
+.idea/**/gradle.xml
+.idea/**/libraries
```
glennmatthews, line 216: This feels like a step backwards to me. We should continue to ignore all of `.idea/` instead of this laundry list of files.

Dockerfile
```diff
@@ -5,6 +5,12 @@ FROM python:${PYTHON_VER}-slim
 RUN pip install --upgrade pip \
   && pip install poetry
 
+RUN apt-get update && apt-get install -y \
+  gcc \
```
glennmatthews, line 9: Adding `gcc` feels pretty heavyweight for this container - what's it needed for?

.gitignore
chadell: I guess it comes from Cookiecutter, we should fix first there :)

glennmatthews: If cruft prevents us from improving things locally as compared to the cookiecutter baseline, that feels like a non-starter to me.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,23 +1,24 @@
 ---
 name: 🐛 Bug Report
-about: Report a reproducible bug in the current release of DiffSync
+about: Report a reproducible bug in the current release of diffsync
```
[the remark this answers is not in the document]
Kircheneer: Agreed. We need to address this at [the source](https://github.com/networktocode-llc/cookiecutter-ntc/blob/main/python/%7B%7Bcookiecutter.project_slug%7D%7D/.github/ISSUE_TEMPLATE/bug_report.md), because currently this is cruft working as intended.

.cruft.json
```diff
@@ -0,0 +1,20 @@
+{
+  "template": "https://github.com/networktocode-llc/cookiecutter-ntc",
+  "commit": "83ae54f27bbeb3ca93326ab1867052ed9d3f6b6a",
```
chadell, line 3: can't we use the "latest" one?

.github/workflows/local.yml
Kircheneer: Unfortunately this doesn't seem to be possible right now in cruft. I agree that this is worse, because now we can't cross-reference between the two sets of workflows. Possibly this is something that we can address by using another paradigm in GitHub actions.

.gitignore
Kircheneer: Agreed. This also needs to be addressed at the template level.

Dockerfile
Kircheneer: This comes from psutil, I believe in conjunction with Buildx:

```
#10 5.989         C compiler or Python headers are not installed on this system. Try to run:
#10 5.989         sudo apt-get install gcc python3-dev
#10 5.989         error: command 'gcc' failed: No such file or directory: 'gcc'
#10 5.989         [end of output]
```

.cruft.json
Kircheneer: Very good point, I was able to use `checkout: main` instead.

.gitignore
Kircheneer: And to your comment Glenn, we can always ignore individual files that we do not want to keep in sync in the `pyproject.toml` file. I do agree that this is rather heavy-handed but since currently we do not have any automatic way of keeping downstream files in sync I believe that this should still be an improvement over the current state, what do you think?

chadell: > If cruft prevents us from improving things locally as compared to the cookiecutter baseline, that feels like a non-starter to me.

My overall approach would be, `cruft` helps to adhere to the cookiecutter. I we do an improvement here (we deviate), `cruft` should raise a warning, and we should focus on change the cookiecutter reference to include the improvement.
I mean, when improving, deviations are expected, and this should help to keep in the right track (not enforcing)

Kircheneer: That is an interesting idea, we will need to hash out the corresponding process further

Kircheneer: To give a quick update here, there is an issue that mentioned the ability to skip specific patches, it is closed however because another part of the issue was implemented. I do believe that this feature specifically wasn't implemented, but I have gone ahead and asked that [here](https://github.com/cruft/cruft/issues/58).

Kircheneer:
Closing this for now, the interesting bits were cherry-picked into #176. Will re-evaluate approach and reconvene later.

PR #168, opened 2022-10-14
Add 'skip' counter to diff.summary()

### LINKED ISSUE #89: Add a new `skip` counter in diff.summary()
### Environment
* DiffSync version:  1.3.0

### Proposed Functionality
Extend the list of counter returned by diff.summary() to include `skip` in addition to `create`, `update`, `delete` & `no-change`

### Use Case
Currently the models that are being skipped because of some global or model flags like `SKIP_UNMATCHED_SRC` | `SKIP_UNMATCHED_DST` are not accounted for in the diff

ubaumann:
Closes #89 

Adding the processed model counter to the Diff object to be able to calculate the number of skipped models.

### DIFF
--- a/diffsync/diff.py
+++ b/diffsync/diff.py
@@ -29,11 +29,11 @@ class Diff:
     def __init__(self):
         """Initialize a new, empty Diff object."""
         self.children = OrderedDefaultDict(dict)
-        self.models_processed = 0
         """DefaultDict for storing DiffElement objects.
 
         `self.children[group][unique_id] == DiffElement(...)`
         """
+        self.models_processed = 0
 
     def __len__(self):
         """Total number of DiffElements stored herein."""
--- a/diffsync/diff.py
+++ b/diffsync/diff.py
@@ -119,8 +119,10 @@ class Diff:
         summary[DiffSyncActions.SKIP] = (
             self.models_processed
             - summary[DiffSyncActions.CREATE]
+            # Updated elements are doubly accumulated in models_processed as they exist in SCR and DST.
             - 2 * summary[DiffSyncActions.UPDATE]
             - summary[DiffSyncActions.DELETE]
+            # 'no-change' elements are doubly accumulated in models_processed as they exist in SCR and DST.
             - 2 * summary["no-change"]
         )
         return summary
--- a/diffsync/diff.py
+++ b/diffsync/diff.py
@@ -29,6 +29,7 @@ class Diff:
     def __init__(self):
         """Initialize a new, empty Diff object."""
         self.children = OrderedDefaultDict(dict)
+        self.models_processed = 0
         """DefaultDict for storing DiffElement objects.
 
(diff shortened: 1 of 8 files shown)


ubaumann:
This implementation only works on the Diff object and not with the DiffElement. At the moment `<Diff>.summary()` returns a dict with the "skip" key and `<DiffElement>.summer()` returns a dict without the "skip" key. So I am not sure if this could let to some confusion.

My first idea was to implement the solution in the DiffElement class, but when a model is skipped, there is no DiffElement returned. Changing this would probably bring more changes, so I decided to first make a simple solution and see what you think about. 

I appreciate feedback and can change the code accordingly.

diffsync/diff.py
```diff
@@ -115,6 +116,13 @@ def summary(self) -> Mapping[Text, int]:
             child_summary = child.summary()
             for key in summary:
                 summary[key] += child_summary[key]
+        summary[DiffSyncActions.SKIP] = (
+            self.models_processed
+            - summary[DiffSyncActions.CREATE]
+            - 2 * summary[DiffSyncActions.UPDATE]
```
chadell, line 122: what is the reason for doubling the update and no.change types?

diffsync/diff.py
```diff
@@ -33,6 +33,7 @@ def __init__(self):
 
         `self.children[group][unique_id] == DiffElement(...)`
         """
+        self.models_processed = 0
```
chadell, line 36: `models` could be a bit confusing, at least to me. What about `objects`?

chadell (commented):
the approach seems reasonable to me, only 2 questions.

diffsync/diff.py
ubaumann: If ther is an action "update" or "no-change" it means the node was in source and destination so the `DiffSyncDiffer` counts it twice. On "create" or "delete" it is only existing in one side.

The model_processed logic was already inplemented in hte DiffSyncDiffer class and I did not want to change the logic of it

diffsync/diff.py
ubaumann: The `DiffSyncDiffer` already had the logic to count the processed models and the name used is `models_prossed`, so I thought it would be good to use the same name for consistency, but I have no problem renaming it
https://github.com/ubaumann/diffsync/blob/689c917428970061346b1715a18575ee0c04bb4d/diffsync/helpers.py#L72

chadell (approved):
LGTM

diffsync/diff.py
Kircheneer: I think it would be good to get a little comment into the code that explains this, because to me it also wasn't immediately clear.

Kircheneer (changes requested):
Looks good to me bar [this](https://github.com/networktocode/diffsync/pull/168/files#r996719231) comment. 🥳

Kircheneer (approved):
LGTM

PR #170, opened 2022-10-18
Add documentation around ordering

itdependsnetworks:


docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
+Given the following Scenario:
+
+```python
+
+class Site(DiffSyncModel):
+    _children = {"vlan": "vlans", "prefix": "prefixes"}
+    [...]
+
+class Device(DiffSyncModel):
+    _children = {"interface": "interfaces"}
```
Kircheneer, line 78: Missing the definition for the `Interface` class here.

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
+- Continue with the next element as defined in the `_children` attribute.
+- Continue with the next element as defined in the `top_level` attribute.
+
+Given the following Scenario:
+
+```python
+
+class Site(DiffSyncModel):
+    _children = {"vlan": "vlans", "prefix": "prefixes"}
+    [...]
```
Kircheneer, line 75: Maybe it would be good to also include the definition of the children fields, i.e. `vlans` and `prefixes`.

Kircheneer: This goes for `Device` and `IPAddress`, too

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
+
+class Vlan(DiffSyncModel):
+    [...]
+
+class Prefix(DiffSyncModel):
+    [...]
+
+class IPAddress(DiffSyncModel):
+    _children = {"ip_address": "ip_addresses"}
```
Kircheneer, line 88: I believe the key should be called `ipaddress` similar to the class, but lower-cased?

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
+```
+
+Would result in processing in the following order:
+
+- site
+- vlan
+- prefix
+- device
+- interface
+- ip_address
```
Kircheneer, line 116: Probably should drop the `_` here, too

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
+- cable
+
+> Note: This applies to the actual diff sync, and not the loading of the data, which is up to the developer to determine the order.
```
Kircheneer, line 119: Perhaps it would be good to correlate this note with the actual names of the methods that do these steps, i.e.

```
> Note: This applies to the actual diff sync (`Diffsync.sync_from/Diffsync.sync_to`), and not the loading of the data (`Diffsync.load`), which is up to the developer to determine the order.
```

Kircheneer (changes requested):
I like this a lot, this answers a question I had in the beginning as well. Since we have dropped Python 3.6 support we are also safe to rely on the fact that dictionaries are ordered now, so that's good.

docs/source/getting_started/01-getting-started.md
itdependsnetworks: That was on purpose, to limit the scope. Can be convinced otherwise.

docs/source/getting_started/01-getting-started.md
itdependsnetworks: I have been keeping the attr names the same as Nautobot orm.

docs/source/getting_started/01-getting-started.md
itdependsnetworks: Much better put, thanks.

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,68 @@ class BackendA(DiffSync):
...
 
+## Model Processing Ordering Logic
+
+The models will be processed in a specfic order as defined by `top_level` atttribute on the `DiffSync` object and then the `_childres` attribute on the `DiffSyncModel`. The processing algorithm can be described as:
+
+- Start with the first element in the model in `top_level`
+- If that model has `_children` set on it, process that first element next, in the order as defined in the dictionary.
+    - Continue processing `_children` as found, until complete end of lineage (e.g. children, children of children, etc.)
+- Continue with the next element as defined in the `_children` attribute.
+- Continue with the next element as defined in the `top_level` attribute.
```
glennmatthews, line 67: As we discussed this morning, if you think of `top_level` and each model's `_children` as forming a tree, the processing order is essentially that of a depth-first traversal of the tree.

docs/source/getting_started/01-getting-started.md
glennmatthews: Shouldn't this be the `_children` of `Interface`, not of `IPAddress` itself?

docs/source/getting_started/01-getting-started.md
```diff
@@ -58,13 +58,12 @@ It's up to the implementer to populate the `DiffSync`'s internal cache with the
 
 ## Model Processing Ordering Logic
 
-The models will be processed in a specfic order as defined by `top_level` atttribute on the `DiffSync` object and then the `_childres` attribute on the `DiffSyncModel`. The processing algorithm can be described as:
+The models will be processed in a specfic order as defined by `top_level` atttribute on the `DiffSync` object and then the `_children` attribute on the `DiffSyncModel`. The processing algorithm is technically a "Preorder Tree Traversal", which means that "a parent node is processed before any of its child nodes is done." This can be described as:
```
Kircheneer, line 61: Who are you quoting here when you quote "a parent node is [...]"?

itdependsnetworks: wikipedia.

Kircheneer: I guess this is largely personal taste but I would drop the quotes here. Feel free to do that or to not do it.

Kircheneer (approved):

docs/source/getting_started/01-getting-started.md
```diff
@@ -56,6 +56,74 @@ class BackendA(DiffSync):
...
 It's up to the implementer to populate the `DiffSync`'s internal cache with the appropriate data. In the example below we are using the `load()` method to populate the cache but it's not mandatory, it could be done differently.
 
+## Model Processing Ordering Logic
+
+The models will be processed in a specfic order as defined by `top_level` atttribute on the `DiffSync` object and then the `_children` attribute on the `DiffSyncModel`. The processing algorithm is technically a "Preorder Tree Traversal", which means that "a parent node is processed before any of its child nodes is done." This can be described as:
+
+- Start with the first element in the model in `top_level`
+- If that model has `_children` set on it, process that first element, in the order as defined in the dictionary.
+- If a child has has `_children` set on it, process that element, and so on until the complete end of lineage (e.g. children, children of children, etc.)
+- Continue with the next element as defined in the `top_level` attribute, and repeat the process
```
glennmatthews, line 66: Perhaps?

```suggestion
- Start with the first element of the first model in `top_level` and process it.
- If that model has `_children` set on it, for each child of each child model, in order:
    - Process that child element.
    - If the child has has `_children` of its own, process its children, and so on until the complete end of lineage (e.g. children, children of children, etc.)
    - Proceed to the next child element, or to the next model in `_children` if done with all elements of that model.
- Repeat for the next element of the top-level model, until done with all elements of that model.
- Continue to the first element of the next model in the `top_level` attribute, and repeat the process, and so on.
```

glennmatthews (approved):

glennmatthews (approved):

PR #174, opened 2022-10-30
Update docker, add methods to load from dictionary, get tree traversa…

itdependsnetworks:
…l, and update docs

* Update docker since it was failing with my build (needed gcc)
* Updated poetry to accommodate other built issues with redislite
* Added method to load from dictionary
* Added method to show the traversal of tree, and docs to go with
* Added method to get one or none

### DIFF
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -466,7 +466,7 @@ class DiffSync:  # pylint: disable=too-many-public-methods
         """Get the initial value order of diffsync object.
 
         Returns:
-            List[str]: List of model-referencing attribute names in the order they are initially processed.
+            List of model-referencing attribute names in the order they are initially processed.
         """
         if hasattr(cls, "top_level") and isinstance(getattr(cls, "top_level"), list):
             value_order = cls.top_level.copy()
@@ -514,7 +514,7 @@ class DiffSync:  # pylint: disable=too-many-public-methods
         """The reverse of `dict` method, taking a dictionary and loading into the inventory.
 
         Args:
-            data (Dict): Dictionary in the format that `dict` would export as
+            data: Dictionary in the format that `dict` would export as
         """
         value_order = self._get_initial_value_order()
         for field_name in value_order:
@@ -672,7 +672,7 @@ class DiffSync:  # pylint: disable=too-many-public-methods
 
     def get_or_none(
         self, obj: Union[Text, DiffSyncModel, Type[DiffSyncModel]], identifier: Union[Text, Mapping]
-    ) -> Union[DiffSyncModel, None]:
+    ) -> Optional[DiffSyncModel]:
         """Get one object from the data store based on its unique id or get a None
 
         Args:
@@ -683,7 +683,7 @@ class DiffSync:  # pylint: disable=too-many-public-methods
             ValueError: if obj is a str and identifier is a dict (can't convert dict into a uid str without a model class)
 
         Returns:
-            DiffSyncModel: DiffSyncModel matching provided criteria
+            DiffSyncModel matching provided criteria
         """
         try:
             return self.get(obj, identifier)
@@ -720,10 +720,10 @@ class DiffSync:  # pylint: disable=too-many-public-methods
         """Get a string describing the tree traversal for the diffsync object.
 
         Args:
-            as_dict (bool): Whether or not to return as a dictionary
+            as_dict: Whether or not to return as a dictionary
 
         Returns:
-            Any: A string or dictionary representation of tree
+            A string or dictionary representation of tree
         """
         value_order = cls._get_initial_value_order()
         output_dict: Dict = {}
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -15,17 +15,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from inspect import isclass
-from typing import Callable, ClassVar, Dict, List, Mapping, Optional, Text, Tuple, Type, Union
+from typing import Any, Callable, ClassVar, Dict, List, Mapping, Optional, Text, Tuple, Type, Union
 
 from pydantic import BaseModel, PrivateAttr
 import structlog  # type: ignore
 
-from .diff import Diff
-from .enum import DiffSyncModelFlags, DiffSyncFlags, DiffSyncStatus
-from .exceptions import DiffClassMismatch, ObjectAlreadyExists, ObjectStoreWrongType, ObjectNotFound
-from .helpers import DiffSyncDiffer, DiffSyncSyncer
-from .store import BaseStore
-from .store.local import LocalStore
+from diffsync.diff import Diff
+from diffsync.enum import DiffSyncModelFlags, DiffSyncFlags, DiffSyncStatus
+from diffsync.exceptions import DiffClassMismatch, ObjectAlreadyExists, ObjectStoreWrongType, ObjectNotFound
+from diffsync.helpers import DiffSyncDiffer, DiffSyncSyncer
+from diffsync.store import BaseStore
+from diffsync.store.local import LocalStore
+from diffsync.utils import get_path, set_key, tree_string
 
 
 class DiffSyncModel(BaseModel):
@@ -396,7 +397,7 @@ class DiffSyncModel(BaseModel):
         childs.remove(child.get_unique_id())
 
 
-class DiffSync:
+class DiffSync:  # pylint: disable=too-many-public-methods
     """Class for storing a group of DiffSyncModel instances and diffing/synchronizing to another DiffSync instance."""
 
     # In any subclass, you would add mapping of names to specific model classes here:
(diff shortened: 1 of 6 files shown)


itdependsnetworks:
Note that mypy does not support recursive typing for now, hence the usage of Any.

diffsync/__init__.py
```diff
@@ -460,6 +461,25 @@ def __len__(self):
         """Total number of elements stored."""
         return self.store.count()
 
+    def _get_value_order(self) -> List[str]:
+        """Get the initial value order of diffsync object.
+
+        Returns:
+            List[str]: List of method names as strings in the order they are initially processed.
```
glennmatthews, line 468: Perhaps?

```suggestion
            List[str]: List of model-referencing attribute names in the order they are initially processed.
```

diffsync/__init__.py
```diff
@@ -489,6 +509,18 @@ def str(self, indent: int = 0) -> str:
...
+        """The reverse of `dict` method, taking a dictionary and loading into the inventory.
+
+        Args:
+            data (Dict): Dictionary in the format that `dict` would export as
+        """
+        value_order = self._get_value_order()
+        for key in value_order:
+            model_obj = getattr(self, key)
+            for values in data.get(key, {}).values():
+                self.add(model_obj(**values))
```
glennmatthews, line 522: Perhaps?
```suggestion
        for field_name in value_order:
            model_class = getattr(self, field_name)
            for values in data.get(field_name, {}).values():
                self.add(model_class(**values))
```

diffsync/__init__.py
```diff
@@ -663,6 +714,34 @@ def get_by_uids(
...
+                set_key(output_dict, [key])
+            if hasattr(model_obj, "_children"):
+                children = getattr(model_obj, "_children")
+                for item in list(children.keys()):
+                    path = get_path(output_dict, key)
+                    if not path:
+                        set_key(output_dict, [key, item])
+                    else:
+                        path.append(item)
+                        set_key(output_dict, path)
```
glennmatthews, line 740: Perhaps?

```suggestion
                for child_key in list(children.keys()):
                    path = get_path(output_dict, key) or [key]
                    path.append(child_key)
                    set_key(output_dict, path)
```

diffsync/__init__.py
```diff
@@ -663,6 +714,34 @@ def get_by_uids(
         """
         return self.store.get_by_uids(uids=uids, model=obj)
 
+    def get_tree_traversal(self, as_dict: bool = False) -> Any:
```
glennmatthews, line 717:
```suggestion
    def get_tree_traversal(self, as_dict: bool = False) -> Union[Text, Mapping]:
```

diffsync/utils.py
```diff
@@ -42,3 +47,49 @@ def __missing__(self, key):
...
+            # i.e. SPACE because LAST, └── , above so no more |
+            yield from _tree(data[path], prefix=prefix + extension)
+
+
+def tree_string(data: dict, root):
+    """String wrapper around `_tree` function to add header and provide tree view of a dictionary."""
+    output = root
+    for line in _tree(data):
+        output = output + "\n" + line
+    return output
```
glennmatthews, line 74: Perhaps?

```suggestion
    return "\n".join([root, *_tree(data)])
```

diffsync/utils.py
```diff
@@ -42,3 +47,49 @@ def __missing__(self, key):
...
+
+
+def set_key(data, keys):
+    """Set a nested dictionary key given a set of keys."""
```
glennmatthews, line 78:
```suggestion
    """Set a nested dictionary key given a list of keys."""
```

diffsync/utils.py
```diff
@@ -42,3 +47,49 @@ def __missing__(self, key):
...
+    current_level = data
+    for item in keys:
+        if item not in current_level:
+            current_level[item] = {}
+        current_level = current_level[item]
```
glennmatthews, line 83:
```suggestion
    for key in keys:
        current_level = current_level.setdefault(key, {})
```

docs/source/getting_started/01-getting-started.md
```diff
@@ -127,6 +127,25 @@ This can be visualized here in the included diagram.
 
 ![Preorder Tree Traversal](../../images/preorder-tree-traversal.drawio.png "Preorder Tree Traversal")
 
+### Mapping Tree Traversal with `get_tree_traversal` method
+
+For your convenience, there is a helper method that will provide a mapping of the order. The `DiffSync.get_tree_traversal` method will return a string, in a tree-like format or optionally a dictionary when setting the `as_dict` method on.
```
glennmatthews, line 132:
```suggestion
For your convenience, there is a helper method that will provide a mapping of the order. The `DiffSync.get_tree_traversal()` method will return a tree-like string, or optionally a dictionary when passing the `as_dict=True` parameter.
```

diffsync/__init__.py
glennmatthews, line 717: Could this be made into a `classmethod` so that you don't have to instantiate the adapter before being able to get this information?

tests/unit/test_diffsync.py
```diff
@@ -898,3 +884,147 @@ class NoDeleteInterfaceDiffSync(BackendA):
     diff = extra_models.diff_from(backend_a)
     print(diff.str())  # for debugging of any failure
     assert not diff.has_diffs()
+
+
+def test_diffsync_tree_traversal(backend_a):
+    assert backend_a.get_tree_traversal(True) == {"site": {"device": {"interface": {}}, "person": {}}, "unused": {}}
+    text = "BackendA\n├── site\n│   ├── device\n│   │   └── interface\n│   └── person\n└── unused"
```
glennmatthews, line 891:
```suggestion
    text = """\
BackendA
├── site
│   ├── device
│   │   └── interface
│   └── person
└── unused"""
```

tests/unit/test_diffsync.py
```diff
@@ -898,3 +884,147 @@ class NoDeleteInterfaceDiffSync(BackendA):
...
+    assert backend_a.get_or_none(backend_a.site, "does-not-exist") is None
+
+
+def test_diffsync_get_value_order(backend_a):
+    assert backend_a._get_value_order() == [  # pylint: disable=protected-access
+        "site",
+        "unused",
+        "device",
+        "interface",
+        "person",
```
glennmatthews, line 1029: This seems like a weird order to me, especially given that it doesn't appear to match the order that `get_tree_traversal()` prints?

itdependsnetworks: Perhaps more descriptive as `_get_initial_value_order`?

diffsync/__init__.py
itdependsnetworks: Let me test, not sure if get_path returns a None or not. Likely just copy this since the workflow is difficult with suggestions to selectively get

diffsync/utils.py
itdependsnetworks: will test locally.

tests/unit/test_diffsync.py
itdependsnetworks: I'm not a fan of either of these views, but I guess lesser of 2 evils.

diffsync/__init__.py
itdependsnetworks: made update, using `cls.__name__` as the first line now.

tests/unit/test_diffsync.py
itdependsnetworks: updated.

diffsync/__init__.py
```diff
@@ -638,6 +670,26 @@ def get(
         """
         return self.store.get(model=obj, identifier=identifier)
 
+    def get_or_none(
+        self, obj: Union[Text, DiffSyncModel, Type[DiffSyncModel]], identifier: Union[Text, Mapping]
+    ) -> Union[DiffSyncModel, None]:
```
Kircheneer, line 675: We can use the `Optional[DiffSyncModel]` here which is equivalent in type

diffsync/utils.py
```diff
@@ -42,3 +47,44 @@ def __missing__(self, key):
         """When trying to access a nonexistent key, initialize the key value based on the internal factory."""
         self[key] = value = self.factory()
         return value
+
+
+# from: https://stackoverflow.com/questions/72618673/list-directory-tree-structure-in-python-from-a-list-of-path-file
+def _tree(data: dict, prefix: str = ""):
```
Kircheneer, line 53: We need to use `typing.Dict` rather than `builtins.dict` as long as we have to support a Python version <3.9

diffsync/utils.py
```diff
@@ -42,3 +47,44 @@ def __missing__(self, key):
...
+    pointers = [TEE] * (len(data) - 1) + [LAST]
+    for pointer, path in zip(pointers, data):
+        yield prefix + pointer + path
+        if isinstance(data[path], dict):  # extend the prefix and recurse:
+            extension = BRANCH if pointer == TEE else SPACE
+            # i.e. SPACE because LAST, └── , above so no more |
+            yield from _tree(data[path], prefix=prefix + extension)
+
+
+def tree_string(data: dict, root):
```
Kircheneer, line 69: We need to use typing.Dict rather than builtins.dict as long as we have to support a Python version <3.9

diffsync/utils.py
```diff
@@ -42,3 +47,44 @@ def __missing__(self, key):
...
+    """String wrapper around `_tree` function to add header and provide tree view of a dictionary."""
+    return "\n".join([root, *_tree(data)])
+
+
+def set_key(data, keys):
+    """Set a nested dictionary key given a list of keys."""
+    current_level = data
+    for key in keys:
+        current_level = current_level.setdefault(key, {})
```
Kircheneer, line 78: Can we get the whole signature type hinted including the return values here? I have opened #175 to make sure we catch this in the CI

Kircheneer (changes requested):
Great stuff! I have a couple type hinting nitpicks but apart from that I like it.

Kircheneer:
> Note that mypy does not support recursive typing for now, hence the usage of Any.

I wasn't able to spot usage of `typing.Any` anywhere except in a function docstring (where the type was not equivalent to that in the signature). I say we just drop new type hints from docstrings (including the ones from this PR) and I make sure I remove all existing ones with #175 ?

Kircheneer (approved):
LGTM

PR #176, opened 2022-11-01
Remove pytest-redislite (closes #139)

### LINKED ISSUE #139: Re-evaluate pytest-redislite as a dependency

Kircheneer:
See #139 for context. Taken out of #144.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,8 @@ jobs:
         uses: "networktocode/gh-action-setup-poetry-environment@v2"
         with:
           python-version: "${{ matrix.python-version }}"
+      - name: "Install redis"
+        run: "sudo apt-get install -y redis"
       - name: "Run poetry Install"
         run: "poetry install"
       - name: "Run Tests"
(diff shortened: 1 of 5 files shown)


chadell (approved):
LGTM

PR #182, opened 2022-11-05
Add methods for get or update instantiate_model

itdependsnetworks:


itdependsnetworks:
Doh’, this doesn’t make sense. When I went to use this, I realized the method was conditionally adding. This is just redundant

itdependsnetworks:
Double Doh'!! ok, red herring, there was an issue, I didn't see where the code was actually issuing a `self.add` (it was nested more), so I thought it was adding.... it was :)

diffsync/store/__init__.py
```diff
@@ -159,6 +159,24 @@ def get_or_instantiate(
 
         return obj, created
 
+    def get_or_instantiate_model(self, obj: "DiffSyncModel") -> Tuple["DiffSyncModel", bool]:
```
Kircheneer, line 162: Maybe this should be called `get_or_add_model_instance`, seeing as
- the model class instance is already instantiated
- the model itself is already known to the `DiffSync` instance at this point (i.e. we aren't "adding" a model)
- "`get_or_add`" also accurately reflects the methods called against `self`

diffsync/store/__init__.py
```diff
@@ -159,6 +159,24 @@ def get_or_instantiate(
...
+        """Attempt to get the object with provided obj identifiers or instantiate obj.
+
+        Args:
+            obj: An obj of the DiffSyncModel to get or create.
+
+        Returns:
+            Provides the existing or new object and whether it was created or not.
```
Kircheneer, line 169: This needs to be updated to reflect the functionality (created vs. instantiated vs. added).

diffsync/store/__init__.py
```diff
@@ -188,6 +206,34 @@ def update_or_instantiate(
...
 
+    def update_or_instantiate_model(self, obj: "DiffSyncModel") -> Tuple["DiffSyncModel", bool]:
+        """Attempt to update an existing object with provided ids/attrs or instantiate obj.
+
+        Args:
+            instance: An instance of the DiffSyncModel to update or create.
+
+        Returns:
+            Provides the existing or new object and whether it was created or not.
+        """
```
Kircheneer, line 217: See comments for current `get_or_instantiate_model` method

diffsync/store/__init__.py
```diff
@@ -188,6 +206,34 @@ def update_or_instantiate(
...
+        model = obj.get_type()
+        ids = obj.get_unique_id()
+        attrs = obj.get_attrs()
+
+        created = False
+        try:
+            obj = self.get(model=model, identifier=ids)
+        except ObjectNotFound:
+            # Add the object to diffsync adapter
```
Kircheneer, line 226: I believe the concept of an adapter is something that the SSoT framework uses, but it is never mentioned in this repository. Possibly:
`# Add the object to the diffsync instance`?

diffsync/store/__init__.py
```diff
@@ -188,6 +206,34 @@ def update_or_instantiate(
...
+            self.add(obj=obj)
+            created = True
+
+        # Update existing obj with attrs
+        for attr, value in attrs.items():
+            if getattr(obj, attr) != value:
```
Kircheneer, line 232: I believe we can omit this check to make it simpler. Just `setattr` always, if its the same the outcome is also the same.

Kircheneer (changes requested):

diffsync/store/__init__.py
```diff
@@ -188,6 +206,33 @@ def update_or_instantiate(
 
         return obj, created
 
+    def update_or_add_model_instance(self, obj: "DiffSyncModel") -> Tuple["DiffSyncModel", bool]:
+        """Attempt to update an existing object with provided ids/attrs or instantiate obj.
+
+        Args:
+            instance: An instance of the DiffSyncModel to update or create.
```
Kircheneer, line 213:
```suggestion
            instance: An instance of the DiffSyncModel to update or add.
```

chadell (approved):
LGTM

Kircheneer (approved):

PR #184, opened 2022-11-08
Resync develop to main after 1.7.0

Kircheneer:


Kircheneer:
I accidentally rebased rather than merged on the release PR merge, I believe this is why it shows the changes here. I believe if we correctly merge this PR we shouldn't have any problems except a slightly mangled git history.

chadell (approved):
LGTM

PR #219, opened 2023-03-17
Type hinting overhaul.

### LINKED ISSUE #175: Type hinting setup overhaul
Currently mypy uses an implicit configuration. We should do the followings things:
- [ ] In what ways do we want to deviate from the default config?
- [ ] Add a `tool.mypy` section to `pyproject.toml`
- [ ] Make sure the configuration is used in the CI
- [ ] Make sure we support Python <3.9 by using `List`, `Dict` etc. from `typing` rather than `list`
- [ ] Drop all type hints from doc strings

Kircheneer:
Closes #175. Things that happened here:
- Fully type hint all function/method definitions
- Active `disallow_untyped_defs` in mypy config in `pyproject.toml`
- Replacing `Mapping` with `Dict`, `Collection` with `List` and `Text` with `str`, because all our mappings are dicts anyway, collections are lists (3.9 could use `dict`/`list`, maybe we can drop 3.8 support with 2.0?) and `Text` is only needed for backwards compatibility with Python 2.x which we aren't offering

### DIFF
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -15,7 +15,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from inspect import isclass
-from typing import Callable, ClassVar, Dict, List, Mapping, Optional, Text, Tuple, Type, Union
+from typing import Callable, ClassVar, Dict, List, Optional, Tuple, Type, Union, Any, Set
+from typing_extensions import Self
 
 from pydantic import BaseModel, PrivateAttr
 import structlog  # type: ignore
@@ -27,6 +28,10 @@ from diffsync.helpers import DiffSyncDiffer, DiffSyncSyncer
 from diffsync.store import BaseStore
 from diffsync.store.local import LocalStore
 from diffsync.utils import get_path, set_key, tree_string
+
+# This workaround is used because we are defining a method called `str` in our class definition, which therefore renders
+# the builtin `str` type unusable.
+StrType = str
 
 
 class DiffSyncModel(BaseModel):
@@ -72,7 +77,7 @@ class DiffSyncModel(BaseModel):
     Note: inclusion in `_attributes` is mutually exclusive from inclusion in `_identifiers`; a field cannot be in both!
     """
 
-    _children: ClassVar[Mapping[str, str]] = {}
+    _children: ClassVar[Dict[str, str]] = {}
     """Optional: dict of `{_modelname: field_name}` entries describing how to store "child" models in this model.
 
     When calculating a Diff or performing a sync, DiffSync will automatically recurse into these child models.
@@ -101,7 +106,7 @@ class DiffSyncModel(BaseModel):
         # Let us have a DiffSync as an instance variable even though DiffSync is not a Pydantic model itself.
         arbitrary_types_allowed = True
 
-    def __init_subclass__(cls):
+    def __init_subclass__(cls) -> None:
         """Validate that the various class attribute declarations correspond to actual instance fields.
 
         Called automatically on subclass declaration.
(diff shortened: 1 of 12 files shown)


diffsync/__init__.py
```diff
@@ -404,13 +404,15 @@ class DiffSync:  # pylint: disable=too-many-public-methods
     # modelname1 = MyModelClass1
     # modelname2 = MyModelClass2
 
-    type: ClassVar[Optional[str]] = None
+    type: Optional[str] = None
```
glennmatthews, line 407: Was this wrong before, or is it a breaking change? I'd thought this *should* be a class variable rather than an instance variable?

diffsync/exceptions.py
```diff
@@ -39,7 +44,7 @@ class ObjectStoreException(Exception):
 class ObjectAlreadyExists(ObjectStoreException):
     """Exception raised when trying to store a DiffSyncModel or DiffElement that is already being stored."""
 
-    def __init__(self, message, existing_object, *args, **kwargs):
+    def __init__(self, message: str, existing_object: Union["DiffSyncModel", "DiffElement"], *args: Any, **kwargs: Any):
```
glennmatthews, line 47: Should `existing_object` just be an `Any`?

pyproject.toml
```diff
@@ -48,6 +48,10 @@ types-toml = "*"
 types-redis = "*"
 pytest-redis = "^2.4.0"
 
+[tool.poetry.group.dev.dependencies]
```
glennmatthews, line 51: Probably should merge the above `tool.poetry.dev-dependencies` into this section?

Kircheneer: Oh yeah I guess my poetry version is too new. We will have to check this for poetry backwards compatibility and leave a note in the docs if its not backwards-compatible.

diffsync/__init__.py
```diff
@@ -132,27 +132,27 @@ def __init_subclass__(cls):
...
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f'{self.get_type()} "{self.get_unique_id()}"'
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self.get_unique_id()
 
-    def dict(self, **kwargs) -> dict:
+    def dict(self, **kwargs: Any) -> Dict:
```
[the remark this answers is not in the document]
Kircheneer: For `**kwargs` you need to annotate the value types as per [PEP 484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values). I went the lazy route here and picked `typing.Any` because I didn't know what exactly the values might be. Need to check this again.

diffsync/__init__.py
Kircheneer: `self.type` was being set in `__init__`, mypy flagged this as a violation.

diffsync/__init__.py
```diff
@@ -132,27 +132,27 @@ def __init_subclass__(cls):
...
+    def json(self, **kwargs: Any) -> str:
         """Convert this DiffSyncModel to a JSON string, excluding the diffsync field by default as it is not serializable."""
         if "exclude" not in kwargs:
             kwargs["exclude"] = {"diffsync"}
         if "exclude_defaults" not in kwargs:
             kwargs["exclude_defaults"] = True
         return super().json(**kwargs)
 
-    def str(self, include_children: bool = True, indent: int = 0) -> str:
+    def to_detailed_string(self, include_children: bool = True, indent: int = 0) -> str:
```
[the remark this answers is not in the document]
Kircheneer: Worked around this.

chadell (approved):
LGTM

PR #220, opened 2023-03-19
Implement DiffSyncModelFlags.NATURAL_DELETION_ORDER.

### LINKED ISSUE #97: DiffSyncModel: Create/Update ordering and Delete ordering
This came up in internal conversations around usage and some potential improvements.

Potentially adding a flag that will set the create/update to process parents before children and then also the reverse, children are deleted before parents.

An example would be Nautobot and the dependencies of objects within Nautobot. Say you want to delete a site, but children objects exist such as devices, you need to delete the devices before deleting the site. This caused some intermittent and hard to troubleshoot scenarios.

It was brought up that some of these deletions would be deferred as well which may not be wanted.

Kircheneer:
Closes #97. Implement `DiffSyncModelFlags.NATURAL_DELETION_ORDER` which handles `DiffSyncAction.DELETE` elements differently by first deleting their children recursively before handling their parents.

## Open points

- [ ] Do we need a global flag for this? If somebody wants this as the default behavior, which is quite a reasonable request to me, they would need to supply this flag over and over again


diffsync/helpers.py
```diff
@@ -350,11 +350,7 @@ def sync_diff_element(self, element: DiffElement, parent_model: Optional["DiffSy
         attrs = diffs.get("+", {})
 
         # Retrieve Source Object to get its flags
-        src_model: Optional["DiffSyncModel"]
-        try:
-            src_model = self.src_diffsync.get(self.model_class, ids)
-        except ObjectNotFound:
-            src_model = None
+        src_model = self.src_diffsync.get_or_none(self.model_class, ids)
```
Kircheneer, line 353: Why is this change here? Pylint was complaining about too many branches and this method was recently introduced and fits this pattern exactly.

diffsync/enum.py
```diff
@@ -47,6 +47,13 @@ class DiffSyncModelFlags(enum.Flag):
     If this flag is set, the model will not be deleted from the target/"to" DiffSync.
     """
 
+    NATURAL_DELETION_ORDER = 0b10000
```
chadell, line 50: to me this looks like the default behavior to be used by default. I can't actually imagine the other behavior use-case. Maybe someone could?

chadell (commented):
to me looks good, but I would make it the default (and sole) behavior

Kircheneer:
> to me looks good, but I would make it the default (and sole) behavior

so @chadell your recommendation would be to drop this into 2.0 and make a backwards-incompatible change, changing the flag to be the default?

chadell:
> > to me looks good, but I would make it the default (and sole) behavior
> 
> so @chadell your recommendation would be to drop this into 2.0 and make a backwards-incompatible change, changing the flag to be the default?

as I see it, this seems more a bugfix than a feature as this deletion mode is the one I would expect by default. Actually, I don't see the other use case. So, my approach would be to do not even define a flag for it.
For the version, I am not sure it deserves a major bump, but if you want to be on the safe side, I'm not against it.

itdependsnetworks:
In agreement with Christian, it seemed odd to me to have this as a flag. In my mind, it should be handled the same way that order children is handled. https://diffsync.readthedocs.io/en/latest/core_engine/02-customize-diff-class.html?#change-the-order-in-which-the-element-are-being-processed

diffsync/enum.py
glennmatthews: It might be a good idea to review known users of this library (https://github.com/topics/diffsync) and confirm their child-deletion behavior. I'm fine with changing the default behavior of DiffSync in a 2.0 version, but if it's relatively low-cost to keep the old behavior as an option, I think it'd be preferable to do so.

docs/source/core_engine/01-flags.md
```diff
@@ -53,13 +53,14 @@ class MyAdapter(DiffSync):
...
+| SKIP_CHILDREN_ON_DELETE | When deleting this model, do not recursively delete its children. Can be used for the case where deletion of a model results in the automatic deletion of all its children.                       | 0b10 |
+| SKIP_UNMATCHED_SRC | Ignore the model if it only exists in the source/"from" DiffSync when determining diffs and syncing. If this flag is set, no new model will be created in the target/"to" DiffSync.               | 0b100 |
+| SKIP_UNMATCHED_DST | Ignore the model if it only exists in the target/"to" DiffSync when determining diffs and syncing. If this flag is set, the model will not be deleted from the target/"to" DiffSync.              | 0b1000 |
+| SKIP_UNMATCHED_BOTH | Convenience value combining both SKIP_UNMATCHED_SRC and SKIP_UNMATCHED_DST into a single flag                                                                                                     | 0b1100 |
+| NATURAL_DELETION_ORDER | When deleting, delete children before instances of this model.                                                                                                                                    | 0b10000 |
```
glennmatthews, line 63: Please verify how this feature interacts with the `SKIP_CHILDREN_ON_DELETE` flag.

Kircheneer: Fixed.

Kircheneer:
My plan is now:
1. Release in 1.8 as a flag
2. Reverse flag (unnatural deletion order) for 2.0
3. When it comes to 3.0, check if anyone is using the reversed flag for unnatural deletion order

glennmatthews (approved):

PR #231, opened 2023-06-20
Simplify model interface

Kircheneer:
This is a draft of an attempt to simplify the interface for defining models a bit. Here is what an example model definition now might look like:

```python
from typing import Annotated, List. Optional

from diffsync.enum import DiffSyncFieldtype
from diffsync import DIffSyncModel


class City(DiffSyncModel):
    _modelname = "city"

    name: Annotated[str, DiffSyncFieldType.IDENTIFIER]
    population: Annotated[int, DiffSyncFieldType.ATTRIBUTE]


class Country(DebugModel):
    _modelname = "country"

    name: Annotated[str, DiffSyncFieldType.IDENTIFIER]
    iso_code: Annotated[Optional[str], DiffSyncFieldType.ATTRIBUTE]
    cities: Annotated[List[City], DiffSyncFieldType.CHILDREN, "city"] = []
```

## Open points

- [ ] Evaluate, whether this breaking change (eyeing at 2.0) is worth it
- [ ] Documentation
- [ ] Think about further possibilites for simplification

### DIFF
--- a/diffsync/__init__.py
+++ b/diffsync/__init__.py
@@ -15,14 +15,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from inspect import isclass
-from typing import Callable, ClassVar, Dict, List, Mapping, Optional, Text, Tuple, Type, Union
+from typing import Callable, ClassVar, Dict, List, Mapping, Optional, Text, Tuple, Type, Union, get_type_hints, get_args
 
 from pydantic import BaseModel, PrivateAttr
 import structlog  # type: ignore
 
 from diffsync.diff import Diff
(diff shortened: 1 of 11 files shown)


itdependsnetworks:
I thought that this was only available in 3.9 and above? https://docs.python.org/3/library/typing.html#typing.Annotated

Kircheneer:
> I thought that this was only available in 3.9 and above? https://docs.python.org/3/library/typing.html#typing.Annotated

True actually, I didn't know. Should use [this](https://typing-extensions.readthedocs.io/en/latest/#Annotated) for compatibility

Kircheneer:
Closing for now.

PR #242, opened 2023-08-24
renames diffsync.DiffSync to diffsync.Adapter

### LINKED ISSUE #233: `diffsync.DiffSync` naming
Since we are calling subclasses of `diffsync.DiffSync`adapters in the examples and even in the docs

> When you initialize the `Diffsync` Adapter class

maybe we should just rename the class to `diffsync.Adapter`? To me this is clearer.

Kircheneer:
Closes #233

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,7 @@ jobs:
       - name: "Check out repository code"
         uses: "actions/checkout@v2"
       - name: "Setup environment"
-        uses: "networktocode/gh-action-setup-poetry-environment@3ea5d3e"
+        uses: "networktocode/gh-action-setup-poetry-environment@3ea5d3ecf382cdcb0c74d4c0ff0629d95fce63c7"
         with:
           python-version: "${{ matrix.python-version }}"
           poetry-version: "${{ matrix.poetry-version }}"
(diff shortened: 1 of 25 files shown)


.github/workflows/ci.yml
```diff
@@ -130,16 +130,18 @@ jobs:
       fail-fast: true
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        poetry-version: ["1.5.1"]
```
chadell, line 133: do you expect to parametrize test by poetry? why not pin it?

.github/workflows/ci.yml
```diff
@@ -130,16 +130,18 @@ jobs:
...
     runs-on: "ubuntu-20.04"
     env:
       PYTHON_VER: "${{ matrix.python-version }}"
     steps:
       - name: "Check out repository code"
         uses: "actions/checkout@v2"
       - name: "Setup environment"
-        uses: "networktocode/gh-action-setup-poetry-environment@v5"
+        uses: "networktocode/gh-action-setup-poetry-environment@3ea5d3e"
```
chadell, line 141: is this a commit? shouldn't we use tags?

chadell (approved):
LGTM, my only question is on the test

Kircheneer:
Sorry @chadell, I was still experimenting with CI breakage, I will set to draft for now

.github/workflows/ci.yml
Kircheneer: This is the act of pinning it, without this its `latest`

.github/workflows/ci.yml
Kircheneer: The latest change that allows me to pin the poetry version is not released

PR #249, opened 2023-11-15
Fixes natural deletion order flag

### LINKED ISSUE #250: Bug with DiffSyncModelFlags.NATURAL_DELETION_ORDER
### Environment
* DiffSync version:  1.9.0

### Observed Behavior

When you have a parent with no changes that has a child with changes and are using `DiffSyncModelFlags.NATURAL_DELETION_ORDER` on the parent, the child will never be synced.

### Expected Behavior

The child should be synced.

### Steps to Reproduce

Check unit test in #249

Kircheneer:
Fixes #250 - see that for more information.

### DIFF
--- a/diffsync/helpers.py
+++ b/diffsync/helpers.py
@@ -369,11 +369,13 @@ class DiffSyncSyncer:  # pylint: disable=too-many-instance-attributes
             natural_deletion_order = bool(dst_model.model_flags & DiffSyncModelFlags.NATURAL_DELETION_ORDER)
             skip_children = bool(dst_model.model_flags & DiffSyncModelFlags.SKIP_CHILDREN_ON_DELETE)
 
+        # Recurse through children to delete if we are supposed to delete the current diff element
         changed = False
         if natural_deletion_order and self.action == DiffSyncActions.DELETE and not skip_children:
             for child in element.get_children():
                 changed |= self.sync_diff_element(child, parent_model=dst_model)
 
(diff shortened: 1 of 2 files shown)


diffsync/helpers.py
```diff
@@ -396,7 +398,7 @@ def sync_diff_element(self, element: DiffElement, parent_model: Optional["DiffSy
 
         self.incr_elements_processed()
 
-        if not natural_deletion_order:
+        if not natural_deletion_order or self.action is None:
```
chadell, line 401: when `self.action` is None?

Kircheneer: I guess it needs a comment - imagine we have a parent and a child and the parent has no changes, then its action will be `None`. This means that without this change the children will not be processed if the parent doesn't have any changes, even if the children do.

chadell: ok, understood

chadell (approved):
LGTM

Kircheneer:
Will wait to see if this correctly addresses the issue by installing from this branch and merge only of it does.

diffsync/helpers.py
glennmatthews: Just comparing against the case above (`if natural_deletion_order and self.action == DiffSyncActions.DELETE and not skip_children:`), should this be the inverse, i.e `if not natural_deletion_order or self.action != DiffSyncActions.DELETE`?

chadell (approved):
LGTM

Issue #256, opened 2023-11-20
Remove `diffsync` field from DiffSyncModel


jamesharr:
### Environment
* DiffSync version: 1.9.0

### Proposed Functionality

- Remove the `diffsync` field from DiffSyncModel
- Add `diffsync` to DiffSyncModel.update and DiffSyncModel.create method signatures

There are a few reasons behind my thinking: 1) It makes the API a bit more consistent 2) It might make it easier for Python to run GC if there's no/fewer cyclical references between objects 3) I didn't see a lot of references to this field outside of create/update/delete

This is likely appropriate for diffsync v2 #232 since it breaks the API.

Other feature requests like #255 may benefit from this change since it would mean there is no assumption that DiffSyncModel has a reference to a given DIffSync

### Use Case

```py
class DiffSyncModel(BaseModel):
    @classmethod
    def create(cls, diffsync: "DiffSync", ids: Dict, attrs: Dict) -> Optional[Self]:
        # Method signature remains the same
        ...

    def update(self, diffsync: "DiffSync", attrs: Dict):
        ...

    def delete(self, diffsync: "DiffSync"):
        ...
```

cardoe:
This actually similarly relates to https://github.com/nautobot/nautobot-app-ssot/issues/272 because of the `job=self` reference which is being passed in many of Nautobot's usage simply to pass a logger down to the model methods of `create`, `update`, `delete`.

What do you ultimately feel the need to pass the `DiffSync` down to each of the models? Since the `DiffSync` holds all the models which are having an operation performed on them?

Kircheneer:
Yeah primarily for me its about having the logging context available.

jamesharr:
> Yeah primarily for me its about having the logging context available.

I don't necessarily need the DiffSync or Logger in the model itself. I need the logger in the CRUD methods and it's also very helpful to have the DiffSync instance in the CRUD methods because that's where we wind up stashing some things like the Nautobot API client, NSO API client, DB connections

cardoe:
But should you? Or should you have a generic `AdapterConfig` object that let's each adapter define the parameters and what gets sent to them. A very rough crappy example... https://github.com/cardoe/diffsync/commit/ffca6ef9fb3af877178d37678c77d57f93fb4321

PR #258, opened 2023-11-20
Remove DiffSyncModel.diffsync field

jamesharr:
This is a draft that would implement #256


jamesharr:
@jdrew82

I'm fine closing this, but I'm wondering about the reasoning. IE were you thinking that a PR wasn't a good place for example code? Or did you decide that this feature would be undesirable

jdrew82:
I think this was a mistake. I was cleaning up stale branches and the next-2.0 branch was one of them. I will say that this PR Is a bit old though and I'm not entirely sure it's still relevant with the changes to DiffSync 2.0 where we replaced the use of diffsync with adapter.

PR #265, opened 2024-02-01
fixes docs build

Kircheneer:


docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
-m2r2==0.2.7
+m2r2==0.3.3.post2
 mistune==0.8.4
-sphinx==4.5.0
+sphinx==6.2.1
 toml==0.10.2
-sphinx-rtd-theme==1.0.0
\ No newline at end of file
+sphinx-rtd-theme==2.0.0
```
glennmatthews, line 5: Should the updates to `m2r2` and `sphinx-rtd-theme` be reflected in `poetry.lock` as well? Or are they already updated there?

Kircheneer: They are already reflected.

glennmatthews (approved):

PR #276, opened 2024-07-17
Chore/update structlog

jamesharr:
datetime.utcnow() is [deprecated in Python 3.12][1], but is still being used in structlog. [Structlog release 23.2.0][2] fixes this. This PR relaxes the structlog version requirement to allow for up to version 24.x, which will get rid of the warning.

Another thing to consider is that DiffSync's use of structlog is pretty basic and structlog's API is pretty stable from the perspective of something generating log messages. Most of the changes pertain to log processors and log configuration, so it might be worth considering removing the max version for structlog. If this is acceptable, I can update the PR to remove the upper version restriction.

[1]: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow
[2]: https://github.com/hynek/structlog/compare/23.1.0...23.2.0


glennmatthews:
Looks like maybe your branch isn't up to date with `develop` and has some merge conflicts as a result. Otherwise, this looks like a very reasonable change. Thanks!

jamesharr:
> Looks like maybe your branch isn't up to date with `develop` and has some merge conflicts as a result. Otherwise, this looks like a very reasonable change. Thanks!

Alright, third time is charm on the rebase. Helps if you rebase off the correct upstream branch

glennmatthews:
Given that structlog uses CalVer instead of SemVer, I'm cautiously OK with removing the upper bound there.

If you don't want to handle all of the changes needed to correspond to updates to linters (`bandit` etc.) then reverting the `poetry.lock` and just doing only `poetry update structlog` should result in a smaller set of diffs there.

glennmatthews (approved):

PR #284, opened 2024-12-10
update pyyaml to 6.0.2 in poetry.lock

gsnider2195:
The pyyaml version in the lock file fails to install due to https://github.com/python-poetry/poetry/issues/8287 / https://github.com/yaml/pyyaml/issues/601


glennmatthews (approved):
Seems reasonable to me. Should we update the dependency in pyproject.toml?

gsnider2195:
> Seems reasonable to me. Should we update the dependency in pyproject.toml?

I think making 6.0.1 or 6.0.2 the minimum would make sense

glennmatthews (approved):

PR #321, opened 2025-12-11
Fix CI

jdrew82:
This PR is to fix the CI release workflow so that the latest release gets pushed to PyPI and GitHub.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,7 @@ jobs:
           path: "dist/"
 
       - name: "Publish package distributions to PyPI"
-        uses: "pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e" # v1.13.0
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           path: "dist/"
 
       - name: "Publish package distributions to PyPI"
-        uses: "pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e"  # v1.13.0
+        uses: "pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e" # v1.13.0
```
smk4664: Diffsync must be different, but we enforce two spaces between content and code comment with yamllint.

smk4664 (approved):
Thanks for making the adjustments!

michalis1 (approved):

Issue #334, opened 2026-02-12
Bug When Creating Diff With Non-Serializable Object


jdrew82:
### Environment
* Python version:  3.10
* DiffSync version:  2.2.0

### Expected Behavior
I expect the Diff to be created without issue.

### Observed Behavior
I observed a TypeError exception being thrown:

```shell
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/celery/app/trace.py", line 479, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/celery/app/trace.py", line 779, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/nautobot/extras/jobs.py", line 1312, in run_job
    result = job(*args, **kwargs)
  File "/usr/local/lib/python3.13/site-packages/nautobot/extras/jobs.py", line 186, in __call__
    return self.run(*args, **deserialized_kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/nautobot/git/itd_nautobot_jobs/jobs/netsuite_sync/data_source.py", line 139, in run
    super().run(*args, **kwargs)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/nautobot_ssot/jobs/base.py", line 640, in run
    self.sync_data(self.memory_profiling)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/nautobot_ssot/jobs/base.py", line 422, in sync_data
    self.load_source_adapter()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/nautobot/git/itd_nautobot_jobs/jobs/netsuite_sync/data_source.py", line 88, in load_source_adapter
    self.source_adapter = JiraAdapter(
                          ~~~~~~~~~~~^
        job=self,
        ^^^^^^^^^
    ...<3 lines>...
        tempo_client=self.tempo_client,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #335, opened 2026-02-12
Fix Deepcopy Bug

jdrew82:
This PR is to fix the issue described in #334.

### DIFF
--- a/changes/334.fixed
+++ b/changes/334.fixed
@@ -1,0 +1,1 @@
+Fix TypeError being thrown when non-serializable object is used in creating a Diff/Adapter.


diffsync/__init__.py
```diff
@@ -484,7 +484,13 @@ def __init_subclass__(cls) -> None:
...
     def __new__(cls, **kwargs):  # type: ignore[no-untyped-def]
         """Document keyword arguments that were used to initialize Adapter."""
-        meta_kwargs = deepcopy(kwargs)
+        meta_kwargs = {}
+        for key, value in kwargs.items():
+            try:
+                meta_kwargs[key] = deepcopy(value)
+            except (TypeError, AttributeError):
+                # Some objects (e.g. Kafka Consumer, DB connections) cannot be deep copied
+                meta_kwargs[key] = value
```
glennmatthews: Do `meta_kwargs` get serialized to JSON at any point (e.g. in nautobot-ssot), and if so, do we need a secondary fix there for non-serializable items in meta_kwargs? Or am I inventing a nonexistent problem?

jdrew82: They are meant to be reused when re-creating the Adapters in SSoT. I haven't quite figured out how to address that particular issue though at this time.

